PR #4, opened 2019-02-13
setusercontext() is in libutil on NetBSD

prlw1:
Code currently assumes that login_cap.h exists if setusercontext is
found so additional check is more of a comment.


wcawijngaards:
Thanks!  Applied this, it also works on FreeBSD, I notice.

PR #6, opened 2019-03-11
Change validity time to a shorter period (<2038)

korhojoa:
7200 days created a certificate with a Not After in 1902. Shortening it to half at least gives you a working certificate.


wcawijngaards:
Thanks for the fix!

PR #10, opened 2019-05-16
Fix memory leaks caused by duplicate rr and include instructions.

sunnypony:


wcawijngaards:
Thank you for the fixes!

PR #12, opened 2019-06-04
Add support for setting send and receive buffer sizes

k0ekk0ek:
This pull request adds support for overriding the default socket send and receive buffer sizes, which is useful in case the default buffer sizes (1m to avoid messages being lost in spikes (I think)) are either too small or too large.

For convenience I've pushed the implementation for forcefully rejecting connections to this branch too, hope that's ok.

### DIFF
--- a/nsd-checkconf.c
+++ b/nsd-checkconf.c
@@ -402,6 +402,8 @@ config_print_zone(nsd_options_type* opt, const char* k, int s, const char *o,
 		SERV_GET_INT(statistics, o);
 		SERV_GET_INT(xfrd_reload_timeout, o);
 		SERV_GET_INT(verbosity, o);
+		SERV_GET_INT(send_buffer_size, o);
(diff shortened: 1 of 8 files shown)


nsd-checkconf.c
```diff
@@ -361,6 +361,8 @@ config_print_zone(nsd_options_type* opt, const char* k, int s, const char *o,
 		/* bin */
 		SERV_GET_BIN(ip_transparent, o);
 		SERV_GET_BIN(ip_freebind, o);
+		SERV_GET_BIN(send_buffer_size, o);
```
k0ekk0ek: This part doesn't seem quite right yet as both `send_buffer_size` and `receive_buffer_size` are integer options as opposed to binary (*yes* or *no*) options. One fix coming up...

wcawijngaards:
Thanks for the PR, it has been merged in a previous commit.

PR #17, opened 2019-06-21
Make configuration parser context aware

k0ekk0ek:
This pull request makes the configuration parser context aware. Previously the parser would see a certain token and update a certain variable in the `nsd_options` struct, which was more-or-less fine since the mapping was mostly 1-1. For some changes that I'm currently working on though it would be easier if the parser is context aware so that when a certain token is read, it knows exactly which block it belongs to. This makes it possible to use e.g. the option `ip-address` in multiple blocks without the need to keep an additional state variable or pointer. The latter is currently used in `master` e.g. to parse `zone:` and `pattern:` blocks (the `current_*` pointers in `cfg_parser`), this request cleans that up too.

The basic tests that Travis does pass, but the checkconf tests do not. I haven't yet traced the cause, but I thought it might be good to submit the first change so that we can (if necessary) discuss the changes somewhat.


k0ekk0ek:
Review comments by @wcawijngaards:

* `block` union in `cfg_parser` is confusing and might cause bugs.
* `c_error` should be annotated with `ATTR_FORMAT`.
* `append_acl` declaration contains invalid whitespace.
* `server:` block is a singleton and multiple blocks should be allowed, as multiple blocks are simply a continuation of the first.
* Type for `number` must not be a `long long`.
* `append_acl` and `append_acl_to_list` are to much alike.

k0ekk0ek:
I've processed the review comments. I found a couple of minor bugs myself that I fixed and ran the checkconf test. It actually finds more errors. Should I update the checkconf.tpkg myself?

wcawijngaards:
Sure. Looks good.

k0ekk0ek:
Updated the checkconf test. Specifically, I added two new tests that verify boolean and number options are handled properly, as they now trigger a parser error (`YYABORT`) and removed invalid boolean and number options from `checkconf.nsd6.conf` so that it now verifies all acls in the configuration. I also verified ordering is intact. If Travis CI passes, I think this is ready to be merged.

k0ekk0ek:
This PR is superseded by PR #37

PR #22, opened 2019-06-29
minimise-any: prefer popular and not large RRset

hdais:
NSD4's RFC8482 (minimizing response for ANY query) just picks one rrset element in linked list with no preference. This can pick a large rrset (e.g. DNSKEY, RRSIG) for response, but this is not suitable for RFC8482 purpose. For example `dig @h.root-servers.net . ANY` often gets all RRSIGs of ".". (H-root adopts NSD4).

This patch introduces simple preference between rtypes on selecting a rrset. It prefers popular and not large rtypes (A, AAAA,..), and lowers large rtypes (DNSKEY, RRSIG,..).

Preferece order is:
1. "Preferred" rtypes (A, AAAA, MX, SOA, PTR)
2. "Normal" rtypes (other than 1 and 3)
3. "Lowered" rtypes (DNSKEY, RRSIG, NSEC)

### DIFF
--- a/query.c
+++ b/query.c
@@ -957,7 +957,16 @@ answer_domain(struct nsd* nsd, struct query *q, answer_type *answer,
 	rrset_type *rrset;
 
 	if (q->qtype == TYPE_ANY) {
-		int added = 0;
+		rrset_type *preferred_rrset = NULL;
+		rrset_type *normal_rrset = NULL;
+		rrset_type *non_preferred_rrset = NULL;
+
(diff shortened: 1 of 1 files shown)


wcawijngaards (approved):
This change looks good to me, for the purpose of making smaller answers.

wcawijngaards:
Thanks for the patch! This looks like it would make even smaller replies for type ANY. I have merged it in the code.

PR #23, opened 2019-07-12
Fix typo in nsd.conf man-page

isbear:


wcawijngaards:
Thank you for the fix!

PR #31, opened 2019-08-05
nsd-control: Add missing stdio header

neheb:
stdio.h is needed in case HAVE_SSL is false as printf gets used.


wcawijngaards:
Thank you for the PR, merged.  Most people have openssl.

PR #32, opened 2019-08-05
tsig: Fix compilation without HAVE_SSL

neheb:
CRYPTO_memcmp is an OpenSSL function that is used outside of the HAVE_SSL path.
This fixes compilation when HAVE_SSL is false.

### DIFF
--- a/tsig.c
+++ b/tsig.c
@@ -19,6 +19,10 @@
 #include "query.h"
 #include "rbtree.h"
 
+#ifndef HAVE_SSL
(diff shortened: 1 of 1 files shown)


wcawijngaards:
Thanks for the patch.  Implemented a fallback routine, because without the fixedtime compare there is a vulnerability.  But no SSL for the hash, anyway.  Also fixed compilation for two other cases where compilation without SSL failed, with ratelimits and the newer DNSoverTLS code.

PR #37, opened 2019-09-26
Make xfrd keep valid zone transfers around until reload succeeds

k0ekk0ek:
This PR prepares NSD for zone verification (previously a separate product called CreDNS) to be merged back. Summary of changes:

* Make configuration options context aware. This allows the same option name to be used for different configuration sections.
* Simplify and update initial socket initialization. This was necessary to have optional IPv6 sockets for both normal operation and verification.
* Set socket options, e.g. reuseaddr and non-blocking operation in separate functions. This makes the code a lot more readable and allows for reuse of the logic when creating the sockets for verification.
* Always use the recvmmsg and sendmmsg code paths. This allows for the elimination of parallel code paths and allows for reuse of the logic for adding handlers.
* Update xfrd to keep zone transfers around until reload succeeds so that updates do not have to fetched again if verification for a single zone fails.

As indicated in the title, it's still a work in progress, but initial review comments have been processed and integrated back into the commits to which they applied.

The last step will be the addition of the code that actually calls the verifier (apart from possibly having to process review comments of course)


tpkg/ixfr_faildel.tdir/ixfr_faildel.test
```diff
@@ -7,8 +7,6 @@
 PRE="../../.."
 DIG=dig
 
-sleep 6
```
wcawijngaards: Maybe best to keep this for slow executing test hosts.

xfrd.c
```diff
@@ -1191,86 +1206,63 @@ xfrd_set_timer(xfrd_zone_type* zone, time_t t)
...
 
-		if(zone->soa_notified_acquired != 0 &&
-			(zone->soa_notified.serial == 0 ||
-		   	compare_serial(ntohl(zone->soa_disk.serial),
-				ntohl(zone->soa_notified.serial)) >= 0))
-		{	/* read was in response to this notification */
-			zone->soa_notified_acquired = 0;
-		}
-		if(zone->soa_notified_acquired && zone->state == xfrd_zone_ok)
+		assert(serial != 0);
```
wcawijngaards: This could be 0 right? In an actual zone transfer, 0 serial is possible in a SOA record.

xfrd.c
```diff
@@ -1191,86 +1206,63 @@ xfrd_set_timer(xfrd_zone_type* zone, time_t t)
...
+		assert(state == xfrd_xfr_corrupt ||
+		       state == xfrd_xfr_inconsistent);
+		while(xfr && 0 != compare_serial(ntohl(xfr->soa.serial),
+		                                 ntohl(serial)))
```
wcawijngaards, line 1214: The call to compare_serial() is not needed, for inequality test.

xfrd.c
```diff
@@ -1965,16 +1986,15 @@ xfrd_parse_received_xfr_packet(xfrd_zone_type* zone, buffer_type* packet,
 				zone->apex_str, zone->master->ip_address_spec));
 			region_destroy(tempregion);
 			return xfrd_packet_bad;
-		}
-		if(zone->soa_disk_acquired != 0 && zone->soa_disk.serial == soa->serial) {
-			DEBUG(DEBUG_XFRD,1, (LOG_INFO, "xfrd: zone %s got "
-						       "update indicating "
-						       "current serial",
+		} else if(compare_serial(ntohl(soa->serial),
+		                         ntohl(last_soa->serial)) == 0)
```
wcawijngaards, line 1976: Also this call to compare_serial() is not needed.  It could be: ntohl(soa->serial) == ntohl(..)

xfrd.c
```diff
@@ -1965,16 +1986,15 @@ xfrd_parse_received_xfr_packet(xfrd_zone_type* zone, buffer_type* packet,
...
+		{
+			DEBUG(DEBUG_XFRD,1, (LOG_INFO,
+				"xfrd: zone %s got update indicating current serial",
 				zone->apex_str));
 			/* (even if notified) the lease on the current soa is renewed */
-			zone->soa_disk_acquired = xfrd_time();
-			if(zone->soa_nsd.serial == soa->serial)
-				zone->soa_nsd_acquired = xfrd_time();
+			*last_soa_acquired = xfrd_time();
+			/* FIXME: zone state should not be updated here(?) */
```
wcawijngaards: Yes the zone state should be updated because we got a new transfer indicating the serial.  Hence we got a new lease time for the serial and thus the timestamp that denotes the leasetime is updated and also the zone state can be changed from (whatever, expired, retrying) to ok.  This means the FIXME comment can be removed, I think?

wcawijngaards: And also the removed line that updates the soa_nsd_acquired if it also needs updating for the timestamp is needed.  Because the zone may be loaded and that zone has a new lease, is more recent, right?

xfrd.c
```diff
@@ -2383,78 +2525,255 @@ find_same_master_notify(xfrd_zone_type* zone, int acl_num_nfy)
...
+			/* transfer must be complete to determine if it is
+			   incremental */
+			if(next_xfr->acquired > 0 && next_xfr->msg_is_ixfr) {
+				xfrd_discard_zone_xfr(zone, next_xfr);
+				next_xfr = xfr->next;
+			} else {
+				break;
 			}
 		}
+		/* refresh if no full transfer was awaiting scheduling */
```
wcawijngaards, line 2629: At this point the comment says you look at if a full transfer was waiting, but the code did not check for that.  It removed ixfrs, but did not check if there is an AXFR or if the loop ended nonNULL.

k0ekk0ek: Correct. I had that in there at some point (comment probably remained), but (from memory) `msg_is_ixfr` might be certain, it didn't seem to be that if that's `false` it's necessarily an axfr. Just doing a refresh in all scenarios seemed like a better decision. Of course, I might have overlooked something or just doing the wrong check all together.

wcawijngaards: It may be required to do a full AXFR after DNSSEC corrupt contents.  Otherwise the master may reply with a smaller incremental update that moves to the failed position every time, it would have to reply with a concatenation or compaction of all updates so that when the zone DNSSEC failure is solved we actually get it.  Probably IXFR is good enough but it depends on the master service of IXFR, right now we rely on the master sending an IXFR update to the latest version, and I think that is actually OK.

xfrd.c
```diff
@@ -2383,78 +2525,255 @@ find_same_master_notify(xfrd_zone_type* zone, int acl_num_nfy)
...
+		xfrd_process_zone_xfrs(zone, committed);
+
+		if(!committed) {
+			continue; /* continue if nothing has been committed */
+		}
+
+		if((zone->soa_notified_acquired != 0 &&
+		    compare_serial(ntohl(zone->soa_notified.serial),
+		                   ntohl(zone->soa_nsd.serial)) <= 0) ||
+		   (zone->soa_notified_acquired < zone->soa_nsd_acquired))
```
wcawijngaards, line 2726: This is a && b || c without brackets, what do you mean?  Also if a notify for a later serial arrives before you process the earlier serial, I would think that soa_notify_acquired (time_t) < nsd_acquired (time_t) but we do not want to just ignore that notification, instead act on it to fetch that serial number.

wcawijngaards: Wait now I see (a&&b) brackets, so the comment is about the time comparison part and ignoring earlier notifications.

xfrd.c
```diff
@@ -2383,78 +2525,255 @@ find_same_master_notify(xfrd_zone_type* zone, int acl_num_nfy)
...
+		}
+
+		/* earlier versions of NSD checked if a transfer was completed
+		   before the reload and concluded that the update must have
+		   failed if the transfer was still around. that shouldn't be
+		   necessary anymore since transfers are now checked
+		   individually. still, the logic is off if a completed
+		   transfer is around after the reload */
+		//assert(zone->soa_disk_acquired &&
+		//       zone->soa_disk_acquired <= xfrd->reload_cmd_last_sent);
```
wcawijngaards: Yes, and I would like an assertion for that.  Also, the `//` operand is imported from C++ and not in early C standards, so I think we should not use that, that improves portability.

tpkg/ixfr_faildel.tdir/ixfr_faildel.test
k0ekk0ek: Fixed.

xfrd.c
k0ekk0ek: Better safe than sorry I guess. Removed the comment. As for "the removed line": `last_soa_acquired` points to either the acquired for the transfer or the one that's loaded. Depending on whether there's a transfer or not.

xfrd.c
k0ekk0ek: Reinstated.

k0ekk0ek:
Creating a new PR, this is getting too complicated :sweat_smile:

PR #39, opened 2019-10-07
Add option and logic to disable additional out-of-zone data.

greg-bock:
NSD follows CNAME, DNAME, and possibly other data outside of the queried zone if the target zone is also configured in NSD. This may not always be desirable. This PR adds a configuration option and logic to disable this behavior. The option defaults to the current behavior. 

nsd.conf
```
server:
        ip-address: 127.0.0.1
        database: ""

remote-control:
        control-enable: yes
        control-interface: 127.0.0.1

zone:
        name: foo.test
        zonefile: /etc/nsd/foo.test.zone
zone:
        name: bar.test
        zonefile: /etc/nsd/bar.test.zone
```

foo.test.zone
```
@ IN SOA ns0.foo.test. username.foo.test. 0 0 0 0 0
cname-zone IN CNAME zone
cname-local IN CNAME  local.bar.test.
cname-remote IN CNAME remote.biz.test.
zone A 127.0.0.1
```
bar.test.zone
```
@ IN SOA ns0.bar.test. username.bar.test. 0 0 0 0 0
local A 127.0.0.2
```

Output:
```
[root@nsd-testing nsd]# dig +short @127.0.0.1 cname-zone.foo.test
zone.foo.test.
127.0.0.1
[root@nsd-testing nsd]# dig +short @127.0.0.1 cname-local.foo.test
local.bar.test.
127.0.0.2
[root@nsd-testing nsd]# dig +short @127.0.0.1 cname-remote.foo.test
remote.biz.test.
```

With the PR changes and additional-from-auth set to no:
```
root@nsd-testing nsd]# dig +short @127.0.0.1 cname-zone.foo.test
zone.foo.test.
127.0.0.1
[root@nsd-testing nsd]# dig +short @127.0.0.1 cname-local.foo.test
local.bar.test.
[root@nsd-testing nsd]# dig +short @127.0.0.1 cname-remote.foo.test
remote.biz.test.
```
I'm open to suggestions on the option name (I took this one from bind although it appears to be obsoleted now). I also contemplated allowing zones/patterns to override the global setting but wasn't sure if calling zone_options_find in the query path was a good idea.

PR #40, opened 2019-10-15
Fix logical typo from confine-to-zone option boolean change.

greg-bock:
Fix logical typo from confine-to-zone option boolean change. I missed this multiple times, sorry I didn't catch it before the merge of #39 .

### DIFF
--- a/query.c
+++ b/query.c
@@ -1236,7 +1236,7 @@ answer_lookup_zone(struct nsd *nsd, struct query *q, answer_type *answer,
 	}
 
 	/*
-	 * If confine-to-zone is set to no do not return additional
+	 * If confine-to-zone is set to yes do not return additional
 	 * information for a zone with a different apex from the query zone.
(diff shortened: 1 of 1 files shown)


greg-bock:
It also appears that I'm not looking at the negative of dname_compare but testing showed it to be working as expected. I'm going to dig into this more.

greg-bock:
dname_compare returns an int not a boolean, fixed the if test. It seems that origzone will always be null if within the same zone (this seems to be why my testing didn't fail)? If so we could probably just get away with just checking that but I'm not familiar enough with the code and other paths/RRs where this might not be the case.

k0ekk0ek:
Thanks @greg-bock. I should've catched that :sweat_smile:. These look fine. I'll be sure to add a test case to verify that it works as expected shortly. Thanks again!

greg-bock:
Thanks again!

PR #43, opened 2019-10-30
Add popen3 and zone resource record iterator

k0ekk0ek:
This PR adds an implementation of `popen3` to execute a process and attach `stdin`, `stdout` and `stderr`. It also extends the suite of unit tests to verify the call works as expected. For easy debugging of unit tests I also extended the `cutest` executable to accept a `-r` parameter to make it possible to run only tests that match the regular expression passed with the `-r` parameter.

Please consider pulling these changes.

### DIFF
--- a/namedb.c
+++ b/namedb.c
@@ -701,3 +701,65 @@ namedb_lookup(struct namedb* db,
 	return domain_table_search(
 		db->domains, dname, closest_match, closest_encloser);
 }
+
+void zone_rr_iter_init(struct zone_rr_iter *iter, struct zone *zone)
+{
+	assert(iter != NULL);
+	assert(zone != NULL);
+	memset(iter, 0, sizeof(*iter));
+	iter->zone = zone;
+}
+
+rr_type *zone_rr_iter_next(struct zone_rr_iter *iter)
+{
+	assert(iter != NULL);
+	assert(iter->zone != NULL);
+
+	if(iter->index == -1) {
+		assert(iter->domain == NULL);
+		assert(iter->rrset == NULL);
+		return NULL;
+	} else if(iter->rrset == NULL) {
+		/* ensure SOA RR is returned first */
+		assert(iter->domain == NULL);
+		assert(iter->index == 0);
+		iter->rrset = iter->zone->soa_rrset;
+	}
+
+	while(iter->rrset != NULL) {
+		if(iter->index < iter->rrset->rr_count) {
+			return &iter->rrset->rrs[iter->index++];
+		}
+		iter->index = 0;
+		if(iter->domain == NULL) {
+			assert(iter->rrset == iter->zone->soa_rrset);
+			iter->domain = iter->zone->apex;
+			iter->rrset = iter->domain->rrsets;
+		} else {
+			iter->rrset = iter->rrset->next;
+		}
+		/* ensure SOA RR is not returned again and RR belongs to zone */
+		while((iter->rrset == NULL && iter->domain != NULL) ||
+		      (iter->rrset != NULL && (iter->rrset == iter->zone->soa_rrset ||
+		                               iter->rrset->zone != iter->zone)))
+		{
+			if(iter->rrset != NULL) {
+				iter->rrset = iter->rrset->next;
+			} else {
+				iter->domain = domain_next(iter->domain);
+				if(iter->domain != NULL &&
+				   dname_is_subdomain(domain_dname(iter->domain),
+				                      domain_dname(iter->zone->apex)))
+				{
+					iter->rrset = iter->domain->rrsets;
+				}
+			}
+		}
+	}
+
+	assert(iter->rrset == NULL);
+	assert(iter->domain == NULL);
+	iter->index = -1;
+
+	return NULL;
+}
--- a/namedb.h
+++ b/namedb.h
@@ -443,4 +443,21 @@ rrset_rrclass(rrset_type* rrset)
 	return rrset->rrs[0].klass;
 }
 
-#endif
+/*
+ * zone_rr_iter can be used to iterate over all RRs in a given zone. the
+ * SOA RRSET is guaranteed to be returned first.
+ */
+typedef struct zone_rr_iter zone_rr_iter_type;
+
+struct zone_rr_iter {
+	zone_type *zone;
+	domain_type *domain;
+	rrset_type *rrset;
+	ssize_t index;
+};
+
+void zone_rr_iter_init(zone_rr_iter_type *iter, zone_type *zone);
+
+rr_type *zone_rr_iter_next(zone_rr_iter_type *iter);
+
+#endif /* _NAMEDB_H_ */
--- a/popen3.c
+++ b/popen3.c
@@ -1,0 +1,172 @@
+#include <assert.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#include "popen3.h"
+
+static void close_pipe(int fds[2])
+{
+	if(fds[0] != -1) {
+		close(fds[0]);
+		fds[0] = -1;
+	}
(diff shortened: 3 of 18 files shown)


k0ekk0ek:
I added a slightly rewritten version of the CreDNS zone resource record iterator and a test that demonstrates it works properly.

k0ekk0ek:
Added a unit test to verify `SIGCHLD` signals are properly captured if handled through libevent as udp/tcp handlers moved to libevent (as opposed to netio) since CreDNS. I actually just used this to see if it works as I expected, but since we have a `libevent` compatible implementation in `mini_event.c`, I figured I might as well leave the test in.

namedb.c
```diff
@@ -701,3 +701,59 @@ namedb_lookup(struct namedb* db,
...
+			assert(iter->rrset == iter->zone->soa_rrset);
+			iter->domain = iter->zone->apex;
+			iter->rrset = iter->domain->rrsets;
+		} else {
+			iter->rrset = iter->rrset->next;
+		}
+		/* ensure SOA RR is not returned again */
+		if(iter->rrset == iter->zone->soa_rrset) {
+			iter->rrset = iter->rrset->next;
+		}
```
wcawijngaards: The rrsets linked list can contain rrsets from other zones.  You have to skip them.  They can exist for the zone apex (parental zone glue, like a DS record), and there can be sub-zones underneath the zone.  Those have to get skipped after the domain_next and dname_is_subdomain checks.  Those zones underneath the parent zone can still also have rrsets for the parent zone, eg. a DS record or glue record addresses from the parent zone.

namedb.h
```diff
@@ -443,4 +443,21 @@ rrset_rrclass(rrset_type* rrset)
 	return rrset->rrs[0].klass;
 }
 
-#endif
+/*
+ * zone_rr_iter can be used to iterate over all RRs in a given zone. the
+ * SOA RRSET is guaranteed to returned first.
```
wcawijngaards: grammatical error in 'to returned'.

popen3.c
```diff
@@ -0,0 +1,170 @@
...
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#include "popen3.h"
+
+static void close_pipe(int fds[2])
+{
+	if(fds[0]) {
```
wcawijngaards: I believe this needs to be compared to -1, not 0.

popen3.c
```diff
@@ -0,0 +1,170 @@
...
+		close(fds[0]);
+		fds[0] = -1;
+	}
+	if(fds[1]) {
```
wcawijngaards: I believe this needs to be compared to -1, not 0.

server.c
```diff
@@ -664,6 +664,22 @@ initialize_dname_compression_tables(struct nsd *nsd)
...
+set_cloexec(struct nsd_socket *sock)
+{
+	assert(sock != NULL);
+
+	if(fcntl(sock->s, F_SETFD, O_CLOEXEC) == -1) {
+		const char *socktype =
+			sock->addr.ai_family == SOCK_DGRAM ? "udp" : "tcp";
+		log_msg(LOG_ERR, "fcntl(..., O_CLOEXEC) failed for %s: %s",
+			socktype, strerror(errno));
+		return -1;
```
wcawijngaards, line 681: The return value is not used, and -1 (instead of 0).

wcawijngaards (approved):
That was the non-test code, and I see no trouble apart from the data from other zones.
Another issue may be that zones could contain out-of-zone glue, eg. a zone example.com contains foo.example.net (note, .net) glue A record.  Not sure if our parser allows that on input, then it may not be an issue.

server.c
k0ekk0ek: In this case that's actually correct. It's a tri-state, not a boolean. The functions return `1` on success, `-1` on error and `0` if it's not implemented.

k0ekk0ek: Or, perhaps you meant the return value for `set_cloexec` itself?

wcawijngaards: Yes the `set_cloexec` return value, you are remarking on `fcntl` above?

namedb.h
k0ekk0ek: Corrected

namedb.c
k0ekk0ek: Corrected. Code actually became somewhat easier to read :+1:. Update the test to include a sub-zone so that the it's guaranteed that they're skipped.

popen3.c
k0ekk0ek: You're right. Corrected.

popen3.c
k0ekk0ek: Right again. Corrected.

server.c
k0ekk0ek: In that case, it's actually correct. It behaves like the other socket option setters. I decided not to check the return value of `set_cloexec` in the function that opens the socket because it's not actually important yet, it's something eventually required for the CreDNS port.

wcawijngaards: Okay, that is good to know.

k0ekk0ek:
clang was giving me some grief. Should be ok now. @wcawijngaards if the build passes, am I cleared to pull this into master?

wcawijngaards:
Sure, I am still going through the test code, but that can be fixed later (if anything is a bug).

k0ekk0ek:
In that case, I'll wait for you to finish. No rush, before the next release is soon enough for me. Thanks

tpkg/cutest/cutest_event.c
```diff
@@ -0,0 +1,364 @@
...
+	}
+	close(fdin[0]);
+	close(fdout[1]);
+	vars->agent.pid = pid;
+	vars->agent.fdin = fdin[1];
+	vars->agent.fdout = fdout[0];
+	event_set(
+		&vars->agent.event,
+		 vars->agent.fdin,
+		 EV_WRITE | EV_CLOSED,
```
wcawijngaards: Does EV_CLOSED really work for other event backends, like select(2) (or with mini_event) ?  Because I guess it may be epoll or kqueue specific.

wcawijngaards (commented):
Wanted to push the ev_closed comment, because it may stop tests to work.

tpkg/cutest/cutest_event.c
```diff
@@ -0,0 +1,364 @@
...
+		vars->children[i].fdout = -1;
+	}
+}
+
+static void
+event_teardown(struct testvars *vars)
+{
+	if(vars->agent.fdin != -1 &&
+	   event_initialized(&vars->agent.event) &&
+	   event_get_fd(&vars->agent.event) == vars->agent.fdin)
```
wcawijngaards, line 244: `event_initialized` and `event_get_fd` are (likely) not supported for other libevent-API that we target.  Eg. minievent, libev, older libevent(likely), and other wrappers. Like EV_CLOSED (just use EV_READ instead, I think, for select and poll), use another variable to indicate initialization (yes this is also a source of bugs, as the initialization gets wrong, you see these kinds of variables in the NSD source, event_added. event_get_fd, I think, by fetching it from the event structure directly, `vars->agent.event.ev_fd`

tpkg/cutest/cutest_event.c
k0ekk0ek: Indeed. Though for select it could be supported, but I've removed it for now.

tpkg/cutest/cutest_event.c
k0ekk0ek: `event_initialized` is supported by libevent 1.4-2.1 and libev. minievent doesn't. I'll implemented that one for minievent. Seems useful to have it there too(?) and read the file descriptor directly from the struct.

k0ekk0ek: On second thought. Initializing the event struct with all zeroes and just checking for equality between `ev_fd` and `fdin` seems like a better idea.

PR #53, opened 2019-11-27
Fix regression with minimal-responses

k0ekk0ek:
This PR fixes NSD behavior wrt to `minimal-repsonses`. I've added a small test case to verify the correct answers are returned with both the option enabled and disabled. It also contains fixes for global static variables not being set for `log-time-ascii` and `round-robin` configuration variables as well some minor cosmetic changes.

Please consider pulling these changes.

### DIFF
--- a/tpkg/minimal_responses.tdir/minimal_responses.test
+++ b/tpkg/minimal_responses.tdir/minimal_responses.test
@@ -1,0 +1,74 @@
+# #-- minimal_responses.test --#
+# source the master var file when it's there
+[ -f ../.tpkg.var.master ] && source ../.tpkg.var.master
+# use .tpkg.var.test for in test variable passing
+[ -f .tpkg.var.test ] && source .tpkg.var.test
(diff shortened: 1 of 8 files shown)


tpkg/minimal_responses.tdir/minimal_responses.test
```diff
@@ -0,0 +1,74 @@
...
+
+diff minimal_responses_no.dig minimal_responses_no.ldns-read-zone
+if [ ${?} -eq 0 ]; then
+  echo "Response with minimal_responses turned off was correct"
+else
+  exit_code=1
+  echo "Response with minimal_responses turned on was incorrect"
+  echo "dig:"
+  cat minimal_responses_no.dig
+  echo "ldsn-read-zone:"
```
wcawijngaards: small spelling error in the test print, s/ldsn/ldns/

wcawijngaards (approved):
Looks good! Like the test.

PR #59, opened 2020-01-11
add FreeBSD support for conf key ip-transparent

buddyns:
This patch implements configuration option "ip-transparent" on FreeBSD (IPv4 and IPv6). Please notice the following:

1. There's some refactoring involved. Different platforms use different macro names for the same socket option, and the previous code block simply repeated multiple times the same `setsockopt()` lines for each different name. This wasn't scalable; this patch uses one linear implementation based on generic macro name, which are resolved to platform-specific socket option names by some pre-processor code upfront. (see comment in code for more details). To add support for a platform with yet a different socket option name, simply define macros the respective macros accordingly.
1. binding both IPv4 and IPv6 addresses is supported on FreeBSD. Linux doesn't seem to support this option on IPv6 programmatically yet, so that's not added. When it is, the pre-processor code makes it easy to add such support: simply add definitions for `NSD_SOCKET_OPTION_TRANSPARENT6` and `NSD_SOCKET_OPTION_TRANSPARENT_OPTLEVEL6` accordingly.
1. `nsd.conf` describes the `ip-freebind` and `ip-transparent` options, seemingly for the same purpose. They are also implemented with same code. This patch only refactors the `ip-transparent` implementation. Admittedly we did not put much effort into reversing what difference :) if equivalent, consider removing one of the options; if not, consider clarifying the difference in the man page.

### DIFF
--- a/server.c
+++ b/server.c
@@ -941,34 +941,64 @@ set_ip_freebind(struct nsd_socket *sock)
 static int
 set_ip_transparent(struct nsd_socket *sock)
 {
+	/*
+	The scandalous preprocessor blob here calls for some explanation :)
+	POSIX does not specify an option to bind non-local IPs, so
+	platforms developed several implementation-specific options,
+	all set in the same way, but with different names.
+	For additional complexity, some platform manage this setting
+	differently for different address families (IPv4 vs IPv6).
+	This scandalous preprocessor blob below abstracts such variability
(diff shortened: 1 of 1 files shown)


wcawijngaards (approved):
The code looks good to me. Thank you for the platform portable code!

PR #61, opened 2020-02-06
Add support for SO_BINDTODEVICE and SO_SETFIB

k0ekk0ek:
This PR adds support for SO_BINDTODEVICE and SO_SETFIB.

### DIFF
--- a/configparser.y
+++ b/configparser.y
@@ -115,6 +115,7 @@ static int parse_range(const char *str, long long *low, long long *high);
 %token VAR_CPU_AFFINITY
 %token VAR_XFRD_CPU_AFFINITY
 %token <llng> VAR_SERVER_CPU_AFFINITY
+%token VAR_DROP_UPDATES
 
 /* dnstap */
 %token VAR_DNSTAP
@@ -237,6 +238,8 @@ server_option:
     { cfg_parser->opt->hide_version = $2; }
   | VAR_HIDE_IDENTITY boolean
     { cfg_parser->opt->hide_identity = $2; }
+  | VAR_DROP_UPDATES boolean
+    { cfg_parser->opt->drop_updates = $2; }
   | VAR_IP4_ONLY boolean
     { if($2) { cfg_parser->opt->do_ip4 = 1; cfg_parser->opt->do_ip6 = 0; } }
   | VAR_IP6_ONLY boolean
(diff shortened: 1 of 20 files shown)


configparser.y
```diff
@@ -475,16 +482,18 @@ server_ranges:
...
         opt->first = (int)first;
         opt->last = (int)last;
-        if(tail) {
-          tail->next = opt;
-          tail = opt;
+        if(servers) {
+          servers = servers->next = opt;
         } else {
-          $$ = tail = opt;
+          servers = cfg_parser->ip->servers = opt;
```
wcawijngaards, line 491: It looks like the assignment to `servers` is unused.  The variable is declared on line 471 and goes out of scope below.

configparser.y
```diff
@@ -226,6 +238,8 @@ server_option:
     { cfg_parser->opt->hide_version = $2; }
   | VAR_HIDE_IDENTITY boolean
     { cfg_parser->opt->hide_identity = $2; }
+  | VAR_DROP_UPDATES boolean
+    { cfg_parser->opt->drop_updates = $2; }
```
wcawijngaards, line 242: You cast (int) some bools above, but not this one.  And not hide_identity. I wonder what is going on.

nsd.c
```diff
@@ -303,28 +303,84 @@ figure_default_sockets(
...
+
+	if(ifa != NULL) {
+		char *colon;
+		size_t len;
+
+		if((colon = strchr(ifa->ifa_name, ':')) != NULL)
+			len = (size_t)((uintmax_t)colon - (uintmax_t)ifa->ifa_name);
+		else
+			len  = strlen(ifa->ifa_name);
+		strlcpy(sock->device, ifa->ifa_name, len);
```
wcawijngaards: You still need to check if the len fits in the size that is available in the `sock->device` buffer.

wcawijngaards (approved):
Reviewed, and the code looks to be nice! Merge it when you want to.

configparser.y
k0ekk0ek: The notation maybe a little icky, but the first time around servers is assigned `cfg_parser->ip->servers` all consecutive operations assign the `next` server to servers. So `servers` is indeed a throwaway variable, but no memory is leaked. The `ip`, of which `servers` is a member is assigned in another block.

configparser.y
k0ekk0ek: True. Removed them. Maybe we should eventually assign the result of `$2 != 0` instead so that the value is guaranteed to be either `1` or `0`, but it's ok for now.

nsd.c
k0ekk0ek: True. Fixed.

k0ekk0ek:
Will be merging this if the build succeeds

PR #63, opened 2020-02-11
Make xfrd keep valid transfers on failed reload

k0ekk0ek:
This PR prepares NSD a bit for zone verification (previously a separate product called CreDNS) to be merged back.

* Update xfrd to keep zone transfers around until reload succeeds so that updates do not have to fetched again if verification for a single zone fails.

### DIFF
--- a/server.c
+++ b/server.c
@@ -1603,7 +1603,8 @@ add_all_soa_to_task(struct nsd* nsd, struct udb_base* taskudb)
 	/* add all SOA INFO to mytask */
 	udb_ptr_init(&task_last, taskudb);
 	for(n=radix_first(nsd->db->zonetree); n; n=radix_next(n)) {
-		task_new_soainfo(taskudb, &task_last, (zone_type*)n->elem, 0);
+		task_new_soainfo(
+			taskudb, &task_last, (zone_type*)n->elem, 0U, xfrd_xfr_new);
 	}
 	udb_ptr_unlink(&task_last, taskudb);
 }
@@ -2189,6 +2190,9 @@ server_reload(struct nsd *nsd, region_type* server_region, netio_type* netio,
 	int ret;
 	udb_ptr last_task;
 	struct sigaction old_sigchld, ign_sigchld;
+	struct zone *zone;
+	struct radnode *node;
+
 	/* ignore SIGCHLD from the previous server_main that used this pid */
 	memset(&ign_sigchld, 0, sizeof(ign_sigchld));
 	ign_sigchld.sa_handler = SIG_IGN;
@@ -2229,6 +2233,18 @@ server_reload(struct nsd *nsd, region_type* server_region, netio_type* netio,
 	server_zonestat_realloc(nsd); /* realloc for new children */
 	server_zonestat_switch(nsd);
 #endif
+
+	for(node = radix_first(nsd->db->zonetree);
+	    node != NULL;
+	    node = radix_next(node))
+	{
+		zone = (zone_type *)node->elem;
+		if(zone->is_updated) {
+			task_new_soainfo(
+				nsd->task[nsd->mytask], &last_task, zone, 0U, xfrd_xfr_ok);
+			zone->is_updated = 0;
+		}
+	}
 
 	/* listen for the signals of failed children again */
 	sigaction(SIGCHLD, &old_sigchld, NULL);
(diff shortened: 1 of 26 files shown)


k0ekk0ek:
@wcawijngaards, @wtoorop, I'll rebase this on latest master after 4.3.0 is released(?)

wtoorop:
ack

k0ekk0ek:
I've added the `ixfr_retry` test to (somewhat) verify behaviour with incorrect transfers is correct, apart from that I worked out one or two kinks that popped up.

k0ekk0ek:
Closing this one because the changes are included in #67.

server.c
```diff
@@ -2402,13 +2418,25 @@ server_main(struct nsd *nsd)
...
+						      "continuing with old "
+						      "database";
+						status = WEXITSTATUS(status);
+					} else {
+						fmt = "Reload process %d "
+						      "exited abnormally, "
+						      "continuing with old "
+						      "database";
+					}
+					log_msg(LOG_WARNING, fmt, child_pid, status);
```
wcawijngaards, line 2436: Could you put the log_msg twice, for the two fmt options. Above in the if statement.  Then also the number of arguments is correct, and checked by the compiler.  I think that is nicer, not that it is wrong now.

tpkg/ixfr_retry.tdir/ixfr_retry.datafile
```diff
@@ -0,0 +1,65 @@
...
+SECTION QUESTION
+baz.foo. IN IXFR
+SECTION ANSWER
+baz.foo. IN SOA    ns.baz.foo. hostmaster.baz.foo. 2 300 4 3000 5
+; deleted items
+baz.foo. IN SOA    ns.baz.foo. hostmaster.baz.foo. 1 3   4 3000 5
+baz.foo. IN SOA    ns.baz.foo. hostmaster.baz.foo. 2 300 4 3000 5
+; added items
+web1     IN A      4.5.6.7
+web2     IN A      5.6.7.8
```
wcawijngaards, line 19: The extension of web1 and web2 is bar.foo, but the IXFR is listed as baz.foo.  See the `$ORIGIN` on line 1.  And on line 24, and the other entries.

tpkg/ixfr_retry.tdir/ixfr_retry.test
```diff
@@ -0,0 +1,62 @@
...
+# requests for baz.foo. before bar.foo. This ensures that changes are
+# processed in the same order. This behavior is used to the advantage of this
+# test because the initial reload must discard the set of changes for the
+# second zone to verify whether the set of changes for the first zone are kept
+# around.
+
+NSD="../../nsd"
+
+#valgrind $NSD -c nsd.conf -L 5 -F 0xFFFF -p $NSD_PORT
+$NSD -c nsd.conf -L 5 -F 0xFFFF -p $NSD_PORT
```
wcawijngaards, line 18: Because this test needs the debug output, you need to test that debug is enabled, and skip the test.  With like `if $NSD -h 2>&1 | grep "L level"; then echo "debug enabled, okay"; else echo "debug output not available, skip test"; exit 0; fi`

tpkg/nodb.tdir/nodb.test
```diff
@@ -17,8 +17,9 @@ echo "export TPKG_PORT=$TPKG_PORT" >> .tpkg.var.test
 echo "export TPKG_NSD_PID=$TPKG_NSD_PID" >> .tpkg.var.test
 echo port: $TPKG_PORT
 
-$TPKG_NSD -c nsd_1.conf -u $LOGNAME -p $TPKG_PORT -P $TPKG_NSD_PID -V 5
+$TPKG_NSD -c nsd_1.conf -u $LOGNAME -p $TPKG_PORT -P $TPKG_NSD_PID -L 5 -F 0xFFFF
 wait_nsd_up nsd.log
+wait_logfile nsd.log "serial 1 is updated to 2" 5
```
wcawijngaards, line 22: It does not look like the -L and -F options are needed by this test, the added wait_logfile does not need it, that is `LOG_INFO` class output.

tpkg/nodb_write.tdir/nodb_write.test
```diff
@@ -17,8 +17,9 @@ echo "export TPKG_PORT=$TPKG_PORT" >> .tpkg.var.test
 echo "export TPKG_NSD_PID=$TPKG_NSD_PID" >> .tpkg.var.test
 echo port: $TPKG_PORT
 
-$TPKG_NSD -c nsd_1.conf -u $LOGNAME -p $TPKG_PORT -P $TPKG_NSD_PID -V 5
+$TPKG_NSD -c nsd_1.conf -u $LOGNAME -p $TPKG_PORT -P $TPKG_NSD_PID -L 5 -F 0xFFFF
 wait_nsd_up nsd.log
+wait_logfile nsd.log "zonefiles write timer (nothing)" 5
```
wcawijngaards, line 22: This test now needs the debug output to be available, like ixfr_retry.tdir test for it?

xfrd.c
```diff
@@ -1171,86 +1190,62 @@ xfrd_set_timer(xfrd_zone_type* zone, time_t t)
...
-	   	compare_serial(ntohl(zone->soa_disk.serial),
-			ntohl(zone->soa_notified.serial)) >= 0))
-	{	/* user provided in response to this notification */
-		zone->soa_notified_acquired = 0;
+	assert(state == xfrd_xfr_ok);
+	/* soa information available, as such the information can be related
+	   to a specific zone transfer (the last zone transfer sent, since soa
+	   information is communicated after verification) and transfer state
+	   must be either xfrd_xfr_ok or xfrd_xfr_drop, though the latter only
+	   occurs if zone verification is actually enabled */
```
wcawijngaards, line 1241: The assert or the comment is wrong, about xfrd_xfr_drop state possible.

PR #79, opened 2020-03-19
Set fib to -1 if no specific addresses are used

k0ekk0ek:
This sets `fib` member to `-1` if no specific addresses are specified and is a FreeBSD specific fix. Mind you, I haven't tested this yet, so please don't merge yet.


wcawijngaards (approved):
Looks good, and compiled on FreeBSD.

k0ekk0ek:
Thanks @wcawijngaards. Yes, it works like a charm, no issues (tested without the first patch), so it's good to go.

Issue #80, opened 2020-03-19
NetBSD and util.c:267:24: warning: implicit declaration of function 'reallocarray'


noloader:
Hi Everyone, @wcawijngaards,

I'm testing nsd master on NetBSD 8.1. One dirty compile shown below. Self tests are OK.

The annoying thing about this report is, manually defining `_OPENBSD_SOURCE` and manually including `<stdlib.h>` early does not clear the issue. It is as if the NetBSD man pages are wrong.

```
gcc -I. -I/usr/pkg/include -g -O2 -flto -c ./util.c
./util.c: In function 'xmallocarray':
./util.c:267:24: warning: implicit declaration of function 'reallocarray' [-Wimplicit-function-declaration]
         void *result = reallocarray(NULL, num, size);
                        ^
./util.c:267:24: warning: initialization makes pointer from integer without a cast [-Wint-conversion]
```

Sorry for the pettiness. Nsd is just about clean on all platforms, so this sticks out like a sore thumb.

-----

According to the NetBSD [`reallocarray(3)` man page](https://netbsd.gw.com/cgi-bin/man-cgi?reallocarray+3+NetBSD-current):

```
#define _OPENBSD_SOURCE
#include <stdlib.h>

void *
reallocarray(void *ptr, size_t nmemb, size_t size);
```

But the configure test result is not used (or `HAVE_REALLOCARRAY` is not used) in `util.c`:

```
$ grep -IR reallocarray util.c
        void *result = reallocarray(NULL, num, size);
                log_msg(LOG_ERR, "reallocarray failed: %s", strerror(errno));
```

Also see this note under *HISTORY* (when in Rome do as the Romans do...):

```
The reallocarray() function first appeared in OpenBSD 5.6.
reallocarray() was redesigned in NetBSD 8 as reallocarr(3).  For compati-
bility reasons it's available since NetBSD 8 in the _OPENBSD_SOURCE
namespace.
```
And:
```
This function is provided for source compatibility with OpenBSD and its
use is discouraged in preference to reallocarr(3).
```
-----
And this may help:

```
$ gcc -dM -E - </dev/null | grep -i -E 'linux|bsd|gnu|i*86|x86|amd' | sort
#define __GNUC_MINOR__ 5
#define __GNUC_PATCHLEVEL__ 0
#define __GNUC_STDC_INLINE__ 1
#define __GNUC__ 5
#define __NetBSD__ 1
#define __amd64 1
#define __amd64__ 1
[...]

PR #81, opened 2020-03-19
Use reallocarr on NetBSD (GH #80)

noloader:
This PR uses `reallocarr` in `xmallocarray` on NetBSD.

NetBSD 8.1 is not finding the declaration of `reallocarray` even with `_OPENBSD_SOURCE` defined. It is causing a noisy compile. Stepping back a bit, the NetBSD [`reallocarray(3)` man page](https://netbsd.gw.com/cgi-bin/man-cgi?reallocarray+3+NetBSD-current) say to use `reallocarr`. This PR follows the NetBSD advice.

According to NetBSD's man page,  `reallocarr` handles overflow properly. There is no need for the `reallocarray` wrapper:

     The reallocarray() function will return NULL if there was overflow or if
     realloc(3) failed setting errno to EOVERFLOW or preserving the value from
     realloc(3).

Also see Issue 80, [util.c:267:24: warning: implicit declaration of function 'reallocarray'](https://github.com/NLnetLabs/nsd/issues/80).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -790,6 +790,25 @@ AC_REPLACE_FUNCS(b64_ntop)
 AC_REPLACE_FUNCS(pselect)
 AC_REPLACE_FUNCS(memmove)
 AC_REPLACE_FUNCS(setproctitle)
+
+AC_MSG_CHECKING([for reallocarr])
+AC_LINK_IFELSE([AC_LANG_SOURCE(AC_INCLUDES_DEFAULT
+[[
+#include <stdlib.h>
+int main(void) {
+	void* p = NULL;
+	int rc;
+	rc = reallocarr(&p, 10, 100);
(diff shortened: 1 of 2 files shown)


wcawijngaards:
Yes this looks good, and #80 then does not need that fix, even though that must be wrong still.  The configure.ac bit could be `AC_CHECK_FUNCS([reallocarr])` instead, which is just shorter for nearly the same.

wcawijngaards:
Please let me know if that AC_CHECK_FUNCS works for you?  Then, if NSDBSD recommends this, we could merge this PR, but since we don't use the 'keep it around on failure', instead free it anyway, that does not really use the reallocarr features.  Having the _OPENBSD_SOURCE define work instead would solve the issues.

Issue #82, opened 2020-03-20
GNU HURD and util.c:1190: warning: sched_setaffinity is not implemented and will always fail


noloader:
Hi Everyone, @wcawijngaards,

Here's the last issue report for HURD. HURD needs a patch for `_GNU_SOURCE` and `<sched.h>`. Then it needs a custom `AC_COMPILE_IFELSE` test. That's attached in the PR.

Now, when linking nsd, we get this message:

```
gcc -g -O2 -flto -L/usr/local/lib -o nsd-checkconf answer.o axfr.o buffer.o conf
iglexer.o configparser.o dname.o dns.o edns.o iterated_hash.o lookup3.o namedb.o
 nsec3.o options.o packet.o query.o rbtree.o radtree.o rdata.o region-allocator.
o rrl.o tsig.o tsig-openssl.o udb.o udbradtree.o udbzone.o util.o bitset.o popen
3.o nsd-checkconf.o cpuset.o strlcat.o strlcpy.o b64_pton.o b64_ntop.o setprocti
tle.o -lcrypto -ldl -levent
/usr/bin/ld: /tmp/nsd.HEkDrr.ltrans3.ltrans.o: in function `set_cpu_affinity':
/home/nsd/util.c:1190: warning: sched_setaffinity is not implemented and will always fail
```

According to [Success in building x86info under GNU/Hurd](https://lists.debian.org/debian-hurd/2011/01/msg00058.html) on the HURD mailing list, this is expected because mach kernels are single core:

> Yes, that's not implemented yet on GNU/Hurd, since the mach is
> mono-processor for now. For the same reason it's actually not a problem
> at all, x86info only needs setaffinity on multi-processor boxes.

The idea is, you have lots of little servers running around and communicating and passing messages rather than a big, monolithic kernel.

Moving forward, `cutest` succeeds:

```
jwalton@hurd-x86:~/nsd$ ./cutest 
srandom(1584675682)
....................................................
OK (52 tests)
SKIP (0 tests)
```

At this point, I know Linux, DragonFly, FreeBSD, NetBSD, OpenBSD and HURD pass the self tests.

`util.c` compiles on Solaris, but not much more. Lex and Yacc are too anemic, and there are no replacements available in `/usr/gnu/bin`. In practice Solaris is not supported.

wcawijngaards:
[...]

PR #83, opened 2020-03-20
Fix GNU HURD sched_setaffinity compile (GH #82)

noloader:
This PR fixes GNU HURD compiles due to `sched_setaffinity`.

GNU HURD was failing compiles due to detection of `sched_setaffinity`. HURD needs `_GNU_SOURCE` defined, and then needs `<sched.h>` included for the proper definition of `cpuset_t`. As such, it needed a custom `AC_COMPILE_IFELSE` test.

We cannot perform `AC_LINK_IFELSE`. See the Issue 82 report for the link warning that could cause mis-detection of `sched_setaffinity`.

There was a lot of trailing whitespace in the files, so `configure.ac` also got `sed -i -e's/[[:space:]]*$//' configure.ac` to trim it.

Also see Issue 82, [util.c:1190: warning: sched_setaffinity is not implemented and will always fail](https://github.com/NLnetLabs/nsd/issues/82).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -112,21 +112,21 @@ AC_DEFINE_UNQUOTED(ZONESDIR, ["`eval echo $zonesdir`"], [NSD default location fo
 
 # default xfrd file location.
 xfrdfile=${dbdir}/xfrd.state
-AC_ARG_WITH([xfrdfile], AC_HELP_STRING([--with-xfrdfile=path], 
+AC_ARG_WITH([xfrdfile], AC_HELP_STRING([--with-xfrdfile=path],
 	[Pathname to the NSD xfrd zone timer state file]), [xfrdfile=$withval])
 AC_DEFINE_UNQUOTED(XFRDFILE, ["`eval echo $xfrdfile`"], [Pathname to the NSD xfrd zone timer state file.])
 AC_SUBST(xfrdfile)
 
 # default zonelist file location.
 zonelistfile=${dbdir}/zone.list
-AC_ARG_WITH([zonelistfile], AC_HELP_STRING([--with-zonelistfile=path], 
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -733,10 +743,43 @@ AS_IF([test x"$ac_cv_type_cpuset_t" = xyes -o x"$ac_cv_type_cpu_set_t" = xyes ],
...
+    #endif
+    #ifdef HAVE_CPUSET_T
+    #define MY_CPUSET_TYPE cpuset_t
+    #endif
+    #ifdef HAVE_CPU_SET_T
+    #define MY_CPUSET_TYPE cpu_set_t
+    #endif
+    #ifdef HAVE_CPUID_T
+    #define MY_CPUSET_TYPE cpuid_t
+    #endif
```
wcawijngaards, line 772: Delete these thee lines, the ifdef HAVE_CPUID_T.  That is for the item in the set, not the set type.

wcawijngaards (approved):
Thanks for the pr, looks good otherwise from a fixup.

wcawijngaards:
Thanks! I merged and fixed the review comment.  I don't think we have previous test for GNU Herd, but it is nice to have support.

configure.ac
wcawijngaards: Fixed in https://github.com/NLnetLabs/nsd/commit/66383945d13eed7c875570956daddaadae116f6e .  Also for the CPU_OR test.

noloader: Perfect, thanks.

noloader:
> I don't think we have previous test for GNU Herd, but it is nice to have support.

Yeah, I'm a big fan of "it just works". Users expect to clone/upack, then `/configure && make && make check && sudo make install`. Anything else just annoys them and generates bug reports and mailing list messages.

PR #72, opened 2020-03-18
Increase Travis testing coverage

noloader:
This PR increases Travis testing coverage.

The Travis changes are the usual suspects. They include GCC and Clang testing on Linux, and Clang testing on OS X. I don't think there's a need for Android or iOS because these are server components. [Corrections, please.]

The first two bullets below covering GCC-9 and Coverity were existing. The tests were retained.

* Retain Coverity GCC-9 test. According to the [Travis folks](https://travis-ci.community/t/run-extended-test-job-if-build-started-from-cron/7625), the CRON condition is `if: env(TRAVIS_EVENT_TYPE) = cron`.
* Retain non-Coverity GCC-9 test.
* Add GCC/Linux, Clang/Linux and Clang OSX tests with no options. Most users will configure like this.
* Add GCC/Linux, Clang/Linux testing using an updated version of Tobias Klien's [Checksec script](https://github.com/slimm609/checksec.sh). It verifies hardening has been applied to an executable.
* Add GCC/Linux, Clang/Linux and Clang OSX UBSan tests.
* Add GCC/Linux, Clang/Linux and Clang OSX ASan tests.
* Add GCC/Linux, Clang/Linux and Clang OSX Analyze tests.
* Add Makefile recipe for `make test`. The recipe also downloads `checksec` and audits the final binaries.

According to the Travis build logs, the current Coverity syntax is deprecated. I think folks are expected to move to the newer style for Coverity. I don't have experience setting up Coverity from Travis so I did not want to change anything.

The builds [test OK on Travis](https://travis-ci.com/github/noloader/nsd/builds/153914506).

PR #84, opened 2020-03-24
Add Cirrus FreeBSD testing

noloader:
This PR adds FreeBSD testing using Cirrus.

NLNetLabs has two actions to perform. First, go to [Cirrus](https://cirrus-ci.com/) and sign up for an account using GitHub credentials. Second, go to GitHub MarketPlace, find the Cirrus app, and install it.

Once things are configured, NLNetLabs will obtain testing results like [here](https://cirrus-ci.com/build/4956744224866304) and shown below.

-----

cirrus-1

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,65 @@
+# https://cirrus-ci.org/examples/
+# https://github.com/curl/curl/blob/master/.cirrus.yml
+# https://github.com/weidai11/cryptopp/blob/master/.cirrus.yml
+
+env:
+  CIRRUS_CLONE_DEPTH: 5
+
+task:
+  matrix:
+    - name: Standard build, FreeBSD 12.1
+      freebsd_instance:
+        image_family: freebsd-12-1
+      pkginstall_script:
+        - pkg update -f
+        - pkg install -y gmake autoconf automake libevent
+      configure_script:
+        - autoconf && autoheader
+        - ./configure --enable-checking
+      compile_script:
+        - make -j 3
+        - make cutest
+      test_script:
+        - ./cutest
+    - name: Standard build, No LTO, FreeBSD 12.1
+      freebsd_instance:
+        image_family: freebsd-12-1
+      pkginstall_script:
+        - pkg update -f
+        - pkg install -y gmake autoconf automake libevent
+      configure_script:
+        - autoconf && autoheader
+        - ./configure --enable-checking --disable-flto
+      compile_script:
+        - make -j 3
+        - make cutest
+      test_script:
+        - ./cutest
+    - name: Standard build, FreeBSD 13.0 (snap)
+      freebsd_instance:
+        image_family: freebsd-13-0-snap
+      pkginstall_script:
+        - pkg update -f
+        - pkg install -y gmake autoconf automake libevent
+      configure_script:
+        - autoconf && autoheader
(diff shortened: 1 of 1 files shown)


k0ekk0ek:
Hi @noloader! Thanks for the PR. I've been thinking additional targets/platforms myself. [Cirrus CI](https://cirrus-ci.org/) seems like a worthwhile addition. Would've been even nicer if we could get builds for some platforms like OpenBSD, NetBSD, etc as well, but most services focus on Linux, Windows and macOS. [SourceHut](https://sourcehut.org/) looks promising too, but I'm not sure it's ready for prime time yet. I was hoping GitHub Actions would eventually support some of the BSD's, but that's [probably not going to happen in the near future](https://github.com/dotnet/runtime/issues/14537). Anyway, before we merge this PR and #72, I'll have to gain some experience with Cirrus CI since NLnet Labs will need to support it. So, I'm going to run some builds and do some tinkering and we'll come back to this after that. Thanks again.

noloader:
@k0ekk0ek,

> I'll have to gain some experience with Cirrus CI since NLnet Labs will need to support it. So, I'm going to run some builds and do some tinkering and we'll come back to this after that. Thanks again.

Ack, take your time.

> Would've been even nicer if we could get builds for some platforms like OpenBSD, NetBSD, etc as well,

+1. I've been looking for something better then manual testing on my VMs.

> SourceHut looks promising too, but I'm not sure it's ready for prime time yet.

Thanks. I'll check them out. You can't have too many tools in the warchest :)

I'm a big fan of testing. For the PR's I have submitted, they are all focused on testing or fixing something testing uncovered. Nearly every bug I fixed in [Crypto++](https://github.com/weidai11/cryptopp) - after the postmortem analysis - was due to a gap in testing.

(My personal feeling is, it is OK to write code with bugs. You can't make an omelet without breaking some eggs. But the trick is, you have to find and fix the bugs before users experience them).

noloader:
@k0ekk0ek ,

Speaking of [PR #72](https://github.com/NLnetLabs/nsd/pull/72/files), the method used in the Coverity test is deprecated. There's a new way to perform Coverity testing detailed at [Travis CI Integration](https://scan.coverity.com/travis_ci) on the Synopsis website.

I don't have experience (yet) with Coverity and Travis. I still perform [manual Coverity testing](https://www.cryptopp.com/wiki/Coverity_Scan).

k0ekk0ek:
Actually, this does the same thing as the script, but better. I analysed what was done and came up with this instead of using the prescribed method. This way, we can just have Coverity Scan on master (I believe that was the main reason) otherwise we'd need a specialized branch.

noloader:
@k0ekk0ek,

I noticed the Coverity jobs fails in non-CRON builds. You may be interested in this from the Travis folks: [Run extended test job if build started from cron?](https://travis-ci.community/t/run-extended-test-job-if-build-started-from-cron/7625/2):

```
    - name: CRON job
      compiler: gcc
      env: TEST_CRON=1
      if: env(TRAVIS_EVENT_TYPE) = cron
```

That should avoid the Coverity job on pull requests and pushes.

noloader:
@k0ekk0ek,

I accidentally closed this. Let me get a new PR ready.

The new PR can take advantage of the Makefile's new audit recipe. The audit recipe runs `checksec` on the binaries produced by a standard build.

Issue #85, opened 2020-03-25
configure.ac and Makefile.in hard codes tools


noloader:
I'm testing NSD master on Solaris 11.3. There are several problems with compile that trace back to lack of precious variables for tools. What happens is, I set a tool like `EGREP=ggrep -E`, but the hardcoded `egrep` is used. There are two problems. First, tools like `grep` and `egrep` are not Posix; and second, Solaris managed to provide tools that are more anemic then Posix.

The fix is threefold. First, [declare the tools as precious in `configure.ac` using `AC_ARG_VAR`](https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Setting-Output-Variables.html):

```
AC_ARG_VAR(SED, [location of the sed program])
AC_ARG_VAR(AWK, [location of the awk program])
AC_ARG_VAR(GREP, [location of the grep program])
...
```

Second, perform the tool tests early:

```
AC_PROG_CC
AC_PROG_SED
AC_PROG_AWK
AC_PROG_GREP
AC_PROG_EGREP
AC_PROG_LN_S
AC_PROG_INSTALL
AC_PROG_LEX
AC_PROG_YACC
```

Third, use the variables in `configure` and `Makefile`. So `Makefile.in` would be changed to:

```
SED 	= @SED@
AWK 	= @AWK@
GREP	= @GREP@
EGREP	= @EGREP@
YACC 	= @YACC@
LEX 	= @LEX@
...

head -`$(EGREP) -n "# Dependencies" $(DEPEND_TARGET) | tail -1 | $(SED) -e 's/:.*$$//'` $(DEPEND_TMP2) > $(DEPEND_TARGET)
```

After the change I was able to get through a compile on Solaris using standard tools sans one error due to an ancient version of OpenSSL.

-----

Making the tools precious variables results in the following when running `./configure --help`:

```
Some influential environment variables:
  SED         location of the sed program
  AWK         location of the awk program
  GREP        location of the grep program
  EGREP       location of the egrep program
  LEX         location of the lex program with GNU extensions (flex)
  YACC        location of the yacc program with GNU extensions (bison)
  CC          C compiler command
  ...
```

-----

Here is a difference in output between Linux and Solaris with precious variables.

**Linux**

```
$ ./configure
...
checking for a sed that does not truncate output... /bin/sed
checking for gawk... gawk
checking for grep that handles long lines and -e... /bin/grep
checking for egrep... /bin/grep -E
checking for flex... flex
...

[...]

PR #86, opened 2020-03-25
Use precious variables for GREP, EGREP, SED, AWK, LEX and YACC

noloader:
This PR uses precious variables for GREP, EGREP, SED, AWK, LEX and YACC. The PR eases building NSD on platforms like Solaris. The PR will make no difference on platforms like Linux.

The precious variables means `configure` will allow the user to override them in the environment, or find the best replacement using `AC_PROG_SED`, `AC_PROG_AWK`, `AC_PROG_GREP` and friends. The variables are then used in the `Makefile.in`.

Sun managed to provide some tools that are more anemic then Posix. Avoiding the anemic tools (like `egrep`) and using the more featured tools (like `ggrep -E`) resolves most build errors on Solaris.

Also see Issue 85, [configure.ac and Makefile.in hard codes tools](https://github.com/NLnetLabs/nsd/issues/85).

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -47,13 +47,17 @@ INSTALL		= $(srcdir)/install-sh -c
 INSTALL_PROGRAM	= $(INSTALL)
 INSTALL_DATA	= $(INSTALL) -m 644
 
-YACC		= @YACC@
+SED 	= @SED@
+AWK 	= @AWK@
+GREP	= @GREP@
+EGREP	= @EGREP@
+YACC 	= @YACC@
 LEX		= @LEX@
 PROTOC_C	= @PROTOC_C@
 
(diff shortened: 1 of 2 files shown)


wcawijngaards (approved):
Review is okay, but want changes (which I committed already).

wcawijngaards:
Thanks for the PR! This is nice to have these tests, I wanted to change two things.  The C compiler want to test that later so that environment can continue to get set up, in case it needs special variables, CFLAGS.  So I moved those lines down a bit.  For the -i test, only if LEX is there, otherwise if it is "" or ":" we can use the flex and bison output included in source tarballs that we distribute for NSD.
Regardless, good to have so merged and included with these fixes.

noloader:
No problems @wcawijngaards. The code keeps improving. That's what matters.

PR #88, opened 2020-03-27
Add README-Travis.md

noloader:
This PR adds README-Travis.md.

README-Travis.md documents what is happening with Travis and Cirrus.

PR #87, opened 2020-03-26
Add Cirrus FreeBSD testing

noloader:
This PR adds Cirrus FreeBSD testing

Also see [PR #84](https://github.com/NLnetLabs/nsd/pull/87). The 84 PR was accidentally closed.

PR #89, opened 2020-03-30
Add Cirrus CI build instructions

k0ekk0ek:
This PR adds build instructions for Cirrus CI and enables builds for Linux, FreeBSD and macOS. I've based my work on the existing `.travis.yml` file already in the repository and PRs #84, #87 and #88 by @noloader. It removes the `checksec` checks because, while useful for package maintainers, they do not add anything for checking PRs against NSD (nobody will use the resulting binaries). Apart from that I've combined the various sanitizer options, thus requiring less builds and added a build for the Coverity Scan service. The Clang 10 build gave a couple of warnings, I've fixed those and added both `/usr/local/opt/libevent` and `/usr/local/opt/openssl` to directories being searched for the respective packages. They're created by [homebrew](https://brew.sh) and thus more users can benefit from them. It also makes the `.cirrus.yml` file consistent across platforms. Last, but not least, I added executing the `tpkg/checkconf` test, so that newly added options make the test fail if they weren't added to the `nsd-checkconf` tool. Oh, and I removed the `.buildkite` directory and `.travis.yml` as we won't be using Buildkite and Cirrus CI checks more boxes than Travis does.


wcawijngaards (approved):
Okay from me.

PR #90, opened 2020-04-03
O_CLOEXEC should be FD_CLOEXEC

phicoh:
F_SETFD takes FD_CLOEXEC

See https://pubs.opengroup.org/onlinepubs/9699919799/functions/fcntl.html


wcawijngaards:
Thanks!  That is correct and a good change.  Merged it into the code base.

PR #91, opened 2020-04-03
nsd-control-setup recreate certificates

gearnode:
The current behavior do not recreate private keys, but the script recreate certificate. I propose to not recreate the certificate (like private key) when certificate already exist.

If you are ok with the patch, i will open the same patch on unbound.

### DIFF
--- a/nsd-control-setup.sh.in
+++ b/nsd-control-setup.sh.in
@@ -5,22 +5,22 @@
 # Copyright (c) 2011, NLnet Labs. All rights reserved.
 #
 # This software is open source.
-# 
+#
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
 # are met:
-# 
+#
 # Redistributions of source code must retain the above copyright notice,
 # this list of conditions and the following disclaimer.
-# 
+#
 # Redistributions in binary form must reproduce the above copyright notice,
 # this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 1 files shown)


wcawijngaards:
Not a bad idea, but the resign is so that it is again valid for $DAYS.  The change would make it so people cannot resign.  Not sure why to have it then?

gearnode:
The intent is to make `nsd-control-setup` idempotent, which allows to run it again and again without to have to `reload` or `restart` NSD. You right with this patch we can't resign and it's an important feature; I add a flag like `-f` or `-r` to resign certificate. Are you ok with this solution?

wcawijngaards:
Sure that would work fine.  Why do you need it to be idempotent, then?  What needs to run this script again and again?  Trying to understand what you need from it, the solution looks fine (or an option for that use case, if that is somehow different from create certificates or resign certificates).

gearnode:
I have idempotent script to bootstrap my servers, and this script exec `nsd-control-setup`, my script is use to create new server from scratch and to update existing server configuration. And to be honest I was surprise when the `nsd-control-setup` resign generated certificates each time it was exec.

gearnode:
I have add `-r` flag to recreate generated certificates.

wcawijngaards (approved):
Looks good.  The -r option is good to have and otherwise it'll create the certificates until they all exist, which is what a lot of startup, or, install, scripts need.

PR #93, opened 2020-04-15
Fix help return code in nsd-control-setup script

k0ekk0ek:
`nsdc-run` test failed because the exit code changed. This "fixes" it.


k0ekk0ek:
Including this in zone-verification PR, closing this one

Issue #18, opened 2019-06-21
crash in nsd ec8d14f8aa32903e894033c678a7268091002a92


jaredmauch:
```Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000055ccdaa1457b in rbtree_next (node=0x74656e0369746c61) at rbtree.c:522
522		if (node->right != RBTREE_NULL) {
Missing separate debuginfos, use: dnf debuginfo-install fstrm-0.4.0-3.fc30.x86_64 glibc-2.29-15.fc30.x86_64 libevent-2.1.8-5.fc30.x86_64 libgcc-9.1.1-1.fc30.x86_64 openssl-libs-1.1.1c-2.fc30.x86_64 protobuf-c-1.3.1-2.fc30.x86_64 zlib-1.2.11-15.fc30.x86_64
(gdb) bt
#0  0x000055ccdaa1457b in rbtree_next (node=0x74656e0369746c61) at rbtree.c:522
#1  0x000055ccdaa088cd in hash_tree_clear (tree=0x55cce8b12740) at namedb.c:301
#2  0x000055ccdaa0a8b9 in nsec3_clear_precompile (db=db@entry=0x55ccdd29c5d0, zone=zone@entry=0x55cce8af3f00) at nsec3.c:406
#3  0x000055ccdaa38b16 in apply_ixfr (softfail=0x7fff8a0f5e60, bytes=<synthetic pointer>, patname=0x7fff8a0f5ea0 "_implicit_svanheule.net", zone_res=<synthetic pointer>, 
    udbz=0x0, rr_count=<synthetic pointer>, delete_mode=<synthetic pointer>, is_axfr=<synthetic pointer>, seq_total=1, seq_nr=0, opt=0x55ccdca7c6d0, serialno=2018111243, 
    zone=0x7fff8a0f66a0 "svanheule.net.", in=0x55cce34c1ef0, db=<optimized out>) at difffile.c:1141
#4  apply_ixfr_for_zone (nsd=nsd@entry=0x55ccdaacac00 <nsd>, zonedb=zonedb@entry=0x55cce8af3f00, in=in@entry=0x55cce34c1ef0, opt=0x55ccdca7c6d0, 
    taskudb=taskudb@entry=0x55ccdd276560, last_task=last_task@entry=0x7fff8a0f87e0, xfrfilenr=2324) at difffile.c:1357
[...]

PR #105, opened 2020-06-10
Clearing hash_tree means just emptying the tree

wtoorop:
Previously (before commit 4ca61188 and/or svn commit 4776) prehashes and corresponding rbtree nodes were part of struct nsec3_domain_data. Clearing the hash_tree would then mean setting the key value of the nodes to NULL to indicate absence of the prehash. But since prehash structs are separately allocated, this is no longer necessary as currently the prehash structs are simply recycled and NULLed.

Currently we have few reports of corrupted hash_tree's during transfer (issue #18 ). Corruption can be caused, either because the prehashed structs of that specific tree were recycled, without them being removed from the tree. Another possibility is that the nodes use memory from recycled pieces of memory that are still in use.

A corrupted transport that we were able to study was for full zone updates (although by IXFR) only. Since this is the most simple recycle path ( with `nsec3_clear_precompile()`) which can hardly go wrong, I currently suspect that the corruption emerged from reusing recycled pieces still in use in other places.  Not writing to the nodes will ease the search for the actual culprit, for example by compiling with CHECK_DOUBLE_FREE checks which might get triggered with the recycling of the prehash structs in . Before this change, writing to key would cause a crash before this check can be made.

In any case, the corruption should be detected when it occurs perhaps testing for rbtree invariants (left->parent == this etc) and we should not crash because of side effects of such corruption.

### DIFF
--- a/namedb.c
+++ b/namedb.c
@@ -301,7 +301,9 @@ void zone_add_domain_in_hash_tree(region_type* region, rbtree_type** tree,
 {
 	if(!*tree)
 		*tree = rbtree_create(region, cmpf);
-	if(node->key) return;
+	if(node->key && node->key == domain
+	&& rbtree_search(*tree, domain) == node)
+		return;
 	memset(node, 0, sizeof(rbnode_type));
 	node->key = domain;
 	rbtree_insert(*tree, node);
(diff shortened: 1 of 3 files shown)


wcawijngaards (approved):
Does not look wrong by itself, and might save time in zone updates, as well as avoid walking the hash tree.

wtoorop:
Yes, but checks are failing. Investigating...

PR #108, opened 2020-06-18
Make the max-retry-time description clearer

nomis:
The word "timeout" implies that nsd will give up after this time. It
doesn't give up but it does increase the retry time exponentially.


wcawijngaards (approved):
Yes that is nice to notice in the manual.

wcawijngaards:
Thanks for the fix!  Merged it, there are likely also upcoming bug fixes to do with the retry timers; in upcoming releases.  But this fixes the description.

PR #109, opened 2020-06-26
Use larger keys for nsd-control in tdir tests

wtoorop:
Because newer OpenSSL fails on the small keys:
error: error in SSL_CTX use_certificate_chain_file crypto error:140AB18F:SSL routines:SSL_CTX_use_certificate:ee key too small


wcawijngaards (approved):
Looks okay to me.  It is only a test so the key size does not matter, but it helps if openssl does not complain about it.

PR #112, opened 2020-07-05
log old and new serials

jaredmauch:
This will log the old and new serial numbers.  Hmm.. the witespace is kinda weird, but I did do this as a patch off the 4.3.1 branch so maybe something changed upstream?


wcawijngaards:
Thanks for the PR!  That looks like a nice printout addition.  Whitespace looks fine to me too.

PR #115, opened 2020-07-23
Fix strlcpy() usage.

millert:
The size argument should be the full size of the buffer, not the number of bytes to copy.  The strlcpy() return value should be checked to verify that truncation did not occur.  From OpenBSD.


wcawijngaards (approved):
Thank you for the patch!  Did not know about those things, even though that was not my commit.  Review looks good, I'll merge it into the code repo.

PR #121, opened 2020-09-03
Increase log level of recreated database from WARNING to ERR

nomis:


wcawijngaards (approved):
This looks good, improves logs.

PR #125, opened 2020-10-05
include config.h in compat/setproctitle.c and fix prototype of `setproctitle`

magenbluten:
- the include guard HAVE_CONFIG_H prevent the inclusion of the actual "config.h"
- setproctitle.c would include <libgen.h> which i am not sure if this is needed, if so config.h should include it?

### DIFF
--- a/compat/setproctitle.c
+++ b/compat/setproctitle.c
@@ -6,9 +6,7 @@
  * See LICENSE for the license.
  *
  */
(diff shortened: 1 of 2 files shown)


wcawijngaards (approved):
Both these changes look good to me.

wcawijngaards:
Thanks for the review fix, merged this into the code base main branch.

PR #126, opened 2020-10-06
minor header hygiene

magenbluten:
- no more recursive inclusion off config.h in cpuset.{c,h}
- remove unnecessary include in options.h
- fix missing shebang in test utils


wcawijngaards:
Thank you for the review fixes, the recursive include is good to remove.

PR #127, opened 2020-10-06
two minor `-Wcast-qual` cleanups

magenbluten:


wcawijngaards:
Thanks fixed.  This fix may please the clang compiler tools.

PR #135, opened 2020-10-14
Feature cookies

magenbluten:
please review @wcawijngaards, @wtoorop 

TODO:
- unit tests
- check on wire that the cookies are propagated to the child processes
- in case the cookie store file is corrupt the first n successfully processed entries overwrite random cookies

### DIFF
--- a/nsd.c
+++ b/nsd.c
@@ -931,6 +931,35 @@ bind8_stats (struct nsd *nsd)
 }
 #endif /* BIND8_STATS */
 
+static
+int cookie_secret_file_read(nsd_type* nsd) {
+	char const* file = nsd->options->cookie_secret_file;
+	assert( file != NULL );
+	FILE* f = fopen(file, "r");
+	/* a non-existing cookie file is not an error */
+	if( f == NULL ) { return errno != EPERM; }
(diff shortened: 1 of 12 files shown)


nsd.c
```diff
@@ -931,6 +931,35 @@ bind8_stats (struct nsd *nsd)
...
+	char const* file = nsd->options->cookie_secret_file;
+	assert( file != NULL );
+	FILE* f = fopen(file, "r");
+	/* a non-existing cookie file is not an error */
+	if( f == NULL ) { return errno != EPERM; }
+	/* cookie secret file exists and is readable */
+	nsd->cookie_count = 0;
+	int corrupt = 0;
+	char secret[NSD_COOKIE_SECRET_SIZE * 2 + 2/*'\n' and '\0'*/];
+	for(size_t count = 0; count < NSD_COOKIE_HISTORY_SIZE; count++ ) {
```
wtoorop: variable declaration inside for loop is very nice (limiting scope), but a c99 feature, so unfortunately it is not portable :(

wtoorop: More occurrences like these above...

wtoorop:
We'll merge in features/cookies first and then review...

PR #141, opened 2020-11-20
ZONEMD RR type

wtoorop:
Support syntax of RR type ZONEMD from draft-ietf-dnsop-dns-zone-digest-14


wcawijngaards:
Yes, it looks fine for syntax of the RR type support!

PR #145, opened 2020-12-02
Add support for connection re-use by secondary for XFR

saradickinson:
*** FOR REVIEW ONLY AT THIS TIME***
* Part of the work to prototype draft-ietf-dprive-xfr-over-tls
* Adds 2 new parameters: `xfrd-conn-reuse` (default no) and `tcp-idle-timeout` (default 10)
* If `xfrd-conn-reuse: yes` the secondary will utilise the existing the mechanisms in place (which are currently only used when `XFRD_MAX_TCP` is exhausted) to re-use connections
* It will additionally leave those connections open and idle for `tcp-idle-timeout`, which is currently a fixed timeout. A later commit will add support for EDNS0 Keepalive, so the server can signal the timeout to use (as described in draft-ietf-dprive-xfr-over-tls)
* Tested against BIND for multiple IXFR and AXFR on the same connection, for the same and different zones
* NOTE: Encountered an issue with NSD reload behaviour and persistent XFR connections when trying to update the `tpkg\long\testplan-axfr.dir` test, will send separate email about this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@
 .cproject
 .project
 .settings/
+
+# Separate build directory
+build/*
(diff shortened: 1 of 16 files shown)


wtoorop:
Instead of the `xfrd-conn-reuse` option, I'd like to have this enabled automatically when connecting over TLS (i.e. XoT) or when a KEEPALIVE EDNS0 option is detected with TCP.
So postponing merge until we have looked into that.

saradickinson:
I think having the default on would be good, I thought an option might be useful so it could be disabled in case of interop problems (e.g a legacy XFR implementation behind a proxy).

wtoorop:
Perhaps it should be configurable per upstream then? For example in the tls-auth: declaration?
It would be good to have tls-cert-bundle configurable per tls-auth: declaration too b.t.w.

PR #149, opened 2020-12-22
Initial implementation of client side XoT.

saradickinson:
- Requires TLS 1.3 support
- Uses a new `tls-auth` clause in the `request-xfr` pattern to specify the authentication domain name to use.


configparser.y
```diff
@@ -796,6 +853,17 @@ pattern_or_zone_option:
...
     }
+  | VAR_REQUEST_XFR VAR_AXFR STRING STRING STRING
+    {
+      acl_options_type *acl = parse_acl_info(cfg_parser->opt->region, $3, $4);
+      acl->use_axfr_only = 1;
+      if(acl->blocked)
+        yyerror("blocked address used for request-xfr");
+      if(acl->rangetype != acl_range_single)
+        yyerror("address range used for request-xfr");
+      acl->tls_auth_name = region_strdup(cfg_parser->opt->region, $4);
```
wcawijngaards: I think this should be `$5` instead of referencing the acl info key name.

```suggestion
      acl->tls_auth_name = region_strdup(cfg_parser->opt->region, $5);
```

saradickinson: Thanks for spotting this!

tpkg/checkconf.tdir/checkconf.check2
```diff
@@ -1,5 +1,5 @@
 start of check
-# Read file checkconf.nsd01.conf: 2 patterns, 2 fixed-zones, 3 keys.
+# Read file checkconf.nsd01.conf: 2 patterns, 2 fixed-zones, 3 key, 0 tls-auth.
```
wcawijngaards: Spelling error, it seems, on this unit test line.

```suggestion
# Read file checkconf.nsd01.conf: 2 patterns, 2 fixed-zones, 3 keys, 0 tls-auth.
```

xfrd-tcp.c
```diff
@@ -24,6 +24,82 @@
...
+{
+	SSL_CTX *ctx;
+	ctx = SSL_CTX_new(TLS_client_method());
+	if (!ctx) {
+		log_msg(LOG_ERR, "xfrd tls: Unable to create SSL ctxt");
+	}
+	else if (SSL_CTX_set_default_verify_paths(ctx) != 1) {
+		SSL_CTX_free(ctx);
+		log_msg(LOG_ERR, "xfrd tls: Unable to set default SSL verify paths");
+	}
```
wcawijngaards: We should not return the deallocated ctx pointer, to stop double free errors.

```suggestion
	else if (SSL_CTX_set_default_verify_paths(ctx) != 1) {
		SSL_CTX_free(ctx);
		log_msg(LOG_ERR, "xfrd tls: Unable to set default SSL verify paths");
		return NULL;
	}
```

xfrd-tcp.c
```diff
@@ -24,6 +24,82 @@
...
+	/* Only trust 1.3 as per the specification */
+	else if (!SSL_CTX_set_min_proto_version(ctx, TLS1_3_VERSION)) {
+		SSL_CTX_free(ctx);
+		log_msg(LOG_ERR, "xfrd tls: Unable to set minimum TLS version 1.3");
+	}
```
wcawijngaards: We should not return the deallocated ctx pointer, to stop double free errors.
```suggestion
	else if (!SSL_CTX_set_min_proto_version(ctx, TLS1_3_VERSION)) {
		SSL_CTX_free(ctx);
		log_msg(LOG_ERR, "xfrd tls: Unable to set minimum TLS version 1.3");
		return NULL;
	}
```

xfrd-tcp.c
```diff
@@ -24,6 +24,82 @@
...
+		log_msg(LOG_ERR, "xfrd tls: Unable to create TLS object");
+		return 0;
+	}
+	SSL_set_connect_state(tp->ssl);
+	(void)SSL_set_mode(tp->ssl, SSL_MODE_AUTO_RETRY);
+	if(!SSL_set_fd(tp->ssl, tp->tcp_w->fd)) {
+		log_msg(LOG_ERR, "xfrd tls: Unable to set TLS fd");
+		SSL_free(tp->ssl);
+		return 0;
+	}
```
wcawijngaards: Set the ssl pointer to NULL like in the other code path before error exit from the function.

```suggestion
	if(!SSL_set_fd(tp->ssl, tp->tcp_w->fd)) {
		log_msg(LOG_ERR, "xfrd tls: Unable to set TLS fd");
		SSL_free(tp->ssl);
		tp->ssl = NULL;
		return 0;
	}
```

xfrd-tcp.c
```diff
@@ -573,6 +662,50 @@ xfrd_tcp_open(struct xfrd_tcp_set* set, struct xfrd_tcp_pipeline* tp,
...
+			err = SSL_get_error(tp->ssl, ret);
+			if (err != SSL_ERROR_WANT_READ && err != SSL_ERROR_WANT_WRITE) {
+				log_msg(LOG_ERR, "xfrd: TLS handshake failed for %s to %s with value: %d", 
+						zone->apex_str, zone->master->ip_address_spec, err);
+				close(fd);
+				xfrd_set_refresh_now(zone);
+				return 0;
+			}
+			/* else wants to be called again */
+		}
```
wcawijngaards, line 701: So I cannot see a lot wrong with this.  You could add if(ret == 0) { close(fd); xfrd_set_refresh_now(zone); return 0; } before the 'else wants to be called again' line.  Then it would be fairly similar to other code that does not really have this issue.

xfrd-tcp.c
```diff
@@ -989,9 +1309,19 @@ xfrd_tcp_pipe_release(struct xfrd_tcp_set* set, struct xfrd_tcp_pipeline* tp,
 		event_del(&tp->handler);
 	tp->handler_added = 0;
 
+#ifdef HACVE_TLS_1_3
```
wcawijngaards: Spelling error in ifdef HAVE_TLS_1_3.

```suggestion
#ifdef HAVE_TLS_1_3
```

xfrd-tcp.c
```diff
@@ -989,9 +1309,19 @@ xfrd_tcp_pipe_release(struct xfrd_tcp_set* set, struct xfrd_tcp_pipeline* tp,
...
+		SSL_shutdown(tp->ssl);
+		SSL_free(tp->ssl);
+		tp->ssl = NULL;
+	}
+#endif
+
 	/* fd in tcp_r and tcp_w is the same, close once */
 	if(tp->tcp_r->fd != -1)
-		close(tp->tcp_r->fd);
+		shutdown(tp->tcp_r->fd, SHUT_RDWR);
```
wcawijngaards: This is the wrong code? It has to close the fd, otherwise it is leaking the file descriptor.  Regardless if you want to call shutdown before close, the close itself has to happen.  Why do you want to call the shutdown? I see no reason for it, and would put the close back again.  Perhaps it was because of the wrong ifdef above on the SSL_shutdown, that a shutdown here gave a nicer result on the other end of the connection, what with not closed but shutdown.

xfrd-tcp.c
```diff
@@ -775,6 +992,104 @@ xfrd_tcp_write(struct xfrd_tcp_pipeline* tp, xfrd_zone_type* zone)
...
+	}
+
+	assert(buffer_remaining(tcp->packet) > 0);
+	ERR_clear_error();
+
+	received = SSL_read(ssl, buffer_current(tcp->packet),
+					buffer_remaining(tcp->packet));
+
+	if (received <= 0) {
+		int err =SSL_get_error(ssl, received);
```
wcawijngaards: Based on other code, you want this space.

```suggestion
		int err = SSL_get_error(ssl, received);
```

xfrd-tcp.c
```diff
@@ -775,6 +992,104 @@ xfrd_tcp_write(struct xfrd_tcp_pipeline* tp, xfrd_zone_type* zone)
...
+	ssize_t received;
+	/* receive leading packet length bytes */
+	if(tcp->total_bytes < sizeof(tcp->msglen)) {
+		ERR_clear_error();
+		received = SSL_read(ssl,
+						(char*) &tcp->msglen + tcp->total_bytes,
+						sizeof(tcp->msglen) - tcp->total_bytes);
+		if (received <= 0) {
+			int err = SSL_get_error(ssl, received);
+			log_msg(LOG_ERR, "ssl_read returned error %d", err);
```
wcawijngaards: I would put the log_msg after the err zero check below, no need to log an error when the connection just closes.  Perhaps on high verbosity or debug, but not for regular operations.

xfrd-tcp.c
```diff
@@ -639,6 +772,80 @@ tcp_conn_ready_for_reading(struct xfrd_tcp* tcp)
...
+static int
+conn_write_ssl(struct xfrd_tcp* tcp, SSL* ssl)
+{
+	ssize_t sent;
+
+	if(tcp->total_bytes < sizeof(tcp->msglen)) {
+		uint16_t sendlen = htons(tcp->msglen);
+		// send
+		int request_length = sizeof(tcp->msglen) - tcp->total_bytes;
+		sent = SSL_write(ssl, (const char*)&sendlen + tcp->total_bytes,
```
wcawijngaards, line 791: ERR_clear_error() is called before you do SSL_read, but not here before SSL_write.

xfrd-tcp.c
```diff
@@ -639,6 +772,80 @@ tcp_conn_ready_for_reading(struct xfrd_tcp* tcp)
...
+			/* incomplete write, resume later */
+			return 0;
+		}
+		assert(tcp->total_bytes >= sizeof(tcp->msglen));
+	}
+
+	assert(tcp->total_bytes < tcp->msglen + sizeof(tcp->msglen));
+
+	int request_length = buffer_remaining(tcp->packet);
+	sent = SSL_write(ssl, buffer_current(tcp->packet), request_length);
```
wcawijngaards, line 823: ERR_clear_error() is called before you call SSL_read, but not here before you call SSL_write.

xfrd-tcp.c
```diff
@@ -639,6 +772,80 @@ tcp_conn_ready_for_reading(struct xfrd_tcp* tcp)
...
+	switch(SSL_get_error(ssl,sent)) {
+		case SSL_ERROR_NONE:
+			if(request_length != sent)
+				log_msg(LOG_ERR, "xfrd: incomplete write of tls");
```
wcawijngaards: It is normal to have incomplete writes because of the packetizing that happens, so it should not be a priority type LOG_ERR to print it during normal operations.  Perhaps debug or high verbosity printout, but it is not an operational issue.

xfrd-tcp.c
```diff
@@ -639,6 +772,80 @@ tcp_conn_ready_for_reading(struct xfrd_tcp* tcp)
...
+	if(tcp->total_bytes < sizeof(tcp->msglen)) {
+		uint16_t sendlen = htons(tcp->msglen);
+		// send
+		int request_length = sizeof(tcp->msglen) - tcp->total_bytes;
+		sent = SSL_write(ssl, (const char*)&sendlen + tcp->total_bytes,
+						 request_length);
+		switch(SSL_get_error(ssl,sent)) {
+			case SSL_ERROR_NONE:
+				if(request_length != sent)
+					log_msg(LOG_ERR, "xfrd: incomplete write of tls");
```
wcawijngaards: It is normal to have incomplete writes because of the packetizing that happens, so it should not be a priority type LOG_ERR to print it during normal operations.  Perhaps debug or high verbosity printout, but it is not an operational issue.

xfrd-tcp.c
```diff
@@ -639,6 +772,80 @@ tcp_conn_ready_for_reading(struct xfrd_tcp* tcp)
...
+				break;
+			default:
+				log_msg(LOG_ERR, "xfrd: generic write problem with tls");
```
wcawijngaards, line 797: It would be nice to print more details, either from errno, or perhaps the function log_crypto_err() for ssl details, depending on what failed.  Perhaps also after other SSL_write and SSL_read and SSL_do_handshake failures.

wcawijngaards (commented):
Code looks good, would want to fix a couple of the issues that I reported (awkward typo that stops SSL_free) before merge, otherwise it looks good once you fixed things.

xfrd-tcp.c
ShivanKaul: Hi Wouter, we initially had close instead of shutdown, but every close caused TCP RSTs. Agreed that we should close the fd... but am trying to figure out how to prevent the RSTs. Perhaps we should do a read before the close?

I've incorporated all the rest of your suggestions except that I'll wait for Sara to weigh in on the open TODO since she found it - will also try to recreate it tomorrow.

wcawijngaards: Yes I would also like to see it work well in practice.  But it sounds like the SSL_shutdown may do the protocol TLS shutdown negotiation, and thus prevent future packets, and thus prevent the TCP RSTs in response to those packets.  And that SSL_shutdown was commented out by the ifdef spelling error in the code above this.  So perhaps with that fix the RSTs are gone?  A read before close does not sound like a solution I have seen elsewhere.

The open issue sounds like something that is a bug, good you can recreate it, that helps figuring out what is going on with the various return states at the time when it goes wrong.

ShivanKaul: Even with the directive correctly formed I get the TCP RSTs.

ShivanKaul: I [asked around](https://stackoverflow.com/questions/66484149/how-do-i-prevent-tcp-resets-after-closing-file-descriptor-that-ive-already-call) and read Bert Hubert's [article](https://blog.netherlabs.nl/articles/2009/01/18/the-ultimate-so_linger-page-or-why-is-my-tcp-not-reliable) on how to handle TCP close gracefully, but even after adding a `read()` and check for 0 before `close(fd)`, I get RSTs... not sure how to proceed.

wcawijngaards: Since NSD is the client here for receiving AXFR data.  If it wants to close, it can close.  Even if there is pending data.  TCP or TLS.  Data that does not arrive is not NSD's problem, because it wants to close.  So it is fine, the other side receiving RST stops sending the AXFR contents, also no problem.  So I do not think the RST is causing an operational problem here.  Just call close().

ShivanKaul:
Hi @wcawijngaards - the git commit history on this PR became very messy with this so I begrudgingly had to cherry pick commits and force push. I believe it should address all your comments, let me know!

wcawijngaards:
You missed the item from xfrd-tcp.c line 677.

The fix commit makes three log_msg calls of type LOG_INFO. That is less then the error level, but it still spams the logs with unused messages during normal operations.  For debug and verbosity messages we have higher output levels defined.  Please move the log at 1065 until after the EOF check below it.  Please move the log at 1016 until after the EOF check below it, it is then fine to use LOG_ERR.  At 749 and 827 use something like `DEBUG(DEBUG_QUERY,2, (LOG_INFO, "text"));` because it is a normal operational thing that can happen for every query (in this case because of packetization of TCP content).  Or just remove those incomplete write log messages.

cesarkuroiwa:
Hello

We managed to do some tests of NSD as XoT client
against our own server implementation. We'd like add a few comments.

We didn't find in the configuration how to set a path for a custom CA,
in order for the TLS handshake to succeed. So we did a small change on
the code to set our custom CA file on the SSL context.

Additionally, we find also that it would be nice if, in the "tls-auth"
section of nsd.conf, there was a possibility to set a path for the
client certificate and key as well, in case the server opts to do
client certificate verification.

saradickinson:
@cesarkuroiwa thanks for the feedback! Agreed - those are two features that are missing and it would be good to add them both to support the mTLS mode of authentication specified in the draft for XoT.

cesarkuroiwa:
If there is interest, I've made a few changes to include some configuration variables and managed to successfully request a XFR from our implemention of a XoT server using mutual TLS authentication by passing a client certificate:

https://github.com/NLnetLabs/nsd/compare/xot-server...cesarkuroiwa:xot-client

tpkg/checkconf.tdir/checkconf.check2
wtoorop: I noticed this is resolved, so resolving conversation.

xfrd-tcp.c
wtoorop: This has been resolved in the current code.

xfrd-tcp.c
wtoorop: Resolved in current code.

xfrd-tcp.c
wtoorop: Resolved in current code.

xfrd-tcp.c
wtoorop: I see this is `close()` now in current code, so resolved.

xfrd-tcp.c
wtoorop: Resolved in current code.

xfrd-tcp.c
wtoorop: I saw the space in the current code, so resolved.

xfrd-tcp.c
wtoorop: It is in the current code, so resolved.

xfrd-tcp.c
wtoorop: Is in the current code, so resolved.

xfrd-tcp.c
wtoorop: Reporting this error is removed from current code, and 0 is returned to try to write again, so resolved.

xfrd-tcp.c
wtoorop: Resolved in current code.

wtoorop:
@cesarkuroiwa What are you using for XoT server to authenticate the client? I think it would be nice to include mutual authentication once we have the server side for that too. I'll keep an eye on your fork :)

The custom CA is nice, but I prefer to use the same name for the option as we use in unbound, which is `tls-cert-bundle`

cesarkuroiwa:
@wtoorop We tested NSD acting as a XoT client against our in-house implementation of a DNS server, which currently act as a hidden master for our zones

I believe the `tls-cert-bundle` approach would be fine, since we would probably use our own private root certificate

xfrd-tcp.c
wtoorop: So this is indeed problematic, because it is spinning on the while loop (causing 100% cpu) and blocking the transfer daemon while doing that. This is resolved in commit 458519a (in PR #183 ) which moves the handshake into the async io based handling of the xfrd_tcp_pipe.

wcawijngaards: That is the best fix! Makes the patch unproblematic for merge, in my opinion.

PR #153, opened 2021-01-28
Repair -fno-common linker errors automatically.

fobser:
Newer C compilers are moving towards -fno-common by default (gcc10, clang11).
This results in linker errors when symbols collide.
nsd contains two yacc parsers, one for the config file, one for zone
files resulting in duplicated lex/yacc symbols.
Instead of trying to repair this with hand-maintained preprocessor
defines use the -P/-p options of flex/yacc to let them do the
renaming for us.


fobser:
I've also tried this on OpenBSD with clang10 with option -fno-common and on fedora33 with clang11.

wcawijngaards:
Thanks for the patch!  That looks like nice options to solve this problem.

Issue #158, opened 2021-03-02
Add Extended DNS Errors (RFC 8914)


TCY16:
It would be nice of NSD had extended errors. In this branch this will be implemented and a PR is expected.

wtoorop:
Feature merged and released

PR #159, opened 2021-03-02
Extended DNS Errors a la RFC8914

wtoorop:
Please don't merge! This is a work in progress!
When finished this resolves issue #158


wtoorop:
TODO items:
- [x] `#define`'s for the EDE OPTION and Error codes (maybe not all but just the ones which make sense for an authoritative server)
- [x] Textual part of EDE (meant for human consumption)
- [x] Find other good places to return an EDE (tip: those places are likely near application of the `RCODE_SET()` macro).
- [x] Find the best spot to determine edns.opt_reserved_space

PR #162, opened 2021-03-09
allow-query: option

wtoorop:
To have ACL for queries.
This is both good for XoT (to prevent discovering the offered zones by sending queries)
and for catalog zones (which should not be query-able)


TCY16 (approved):

query.c
```diff
@@ -1297,6 +1297,42 @@ answer_lookup_zone(struct nsd *nsd, struct query *q, answer_type *answer,
...
+			DEBUG(DEBUG_QUERY,1, (LOG_INFO, "query %s passed acl %s %s",
+				dname_to_string(q->qname, NULL),
+				why->ip_address_spec,
+				why->nokey?"NOKEY":
+				(why->blocked?"BLOCKED":why->key_name)));
+		} else { 
+			if (verbosity >= 2) {
+				char address[128];
+				addr2str(&q->addr, address, sizeof(address));
+				VERBOSITY(2, (LOG_INFO, "query %s from %s refused, %s%s",
```
wcawijngaards: I thought you would want a space between the two explanatory strings.

```suggestion
				VERBOSITY(2, (LOG_INFO, "query %s from %s refused, %s %s",
```

wcawijngaards (approved):
Looks good, if it is useful for people we can add it.

Issue #154, opened 2021-02-04
TXT with parentheses fails in 4.3.5


sthen:
The fix for #150 breaks records split between multiple lines with parentheses, e.g.

```
test TXT (
  "abcdef"
  "ghijk")
```

Either resulting in just "syntax error", or "nested parentheses" and "syntax error" (I think the latter when there are multiple such records in a zone).

The syntax is used very often with DKIM keys.

bes-internal:
NSD starts with no errors but return old zone (is it cached?) and no return txt records with parentheses and multiline at all!

`nsd-checkzone` is not very helpful in finding the reason, because it just gives out uninformative "syntax error" on next line number

wtoorop:
This is resolved with the merge of PR #164 which also contains a -p option for nsd-checkzone to print the successfully read zone (and aid in debugging these kind of issues).

PR #164, opened 2021-03-11
Bugfix/issue 150 and 154

### LINKED ISSUE #150: TXT record validation difference with BIND
I stumbled upon a slight difference in zonfile validation with `nsd-checkzone` compared to the `named-checkzone` from bind9.

Assuming this TXT record in a zonfile:

```
foo       IN  TXT   "a"b
```
Please note, that there is no whitespace between `"a"` and `b`.

* `named-checkzone` considers a zonefile with such a TXT record valid. If loaded into bind, the zonefile is accepted and responds for the record.
* `nsd-checkzone` exits with a syntax error, having such a record in a zonefile also prevents NSD from starting. Only if you add a withespace between `"a"` and `b`, `nsd-checkzone` accepts this.
* `kzonecheck` (knot dns server) has the same behaviour as nsd on this topic.

I could not find the section in any RFC that covers this, but I think zonefiles should be cross compatible with bind providing the reference.

wtoorop:
Resolve issue #154 and partially resolve #150 

For reference, given this zone:
```
$ORIGIN example.
$TTL    1
@       SOA     ns admin 1 7200 3600 1209600 3600
        ns      @
        A       192.0.2.1
        TXT     "ab"cd ef"gh" "ij""kl"
test    TXT     (
  "abcdef"
  "ghijk")
```

BIND prints this as:
```
$ named-checkzone -o - example example 
zone example/IN: loaded serial 1
example.	      1	IN SOA		ns.example. admin.example. 1 7200 3600 1209600 3600
example.	      1	IN NS		example.
example.	      1	IN A		192.0.2.1
example.	      1	IN TXT		"ab" "cd" "ef" "gh" "ij" "kl"
test.example.	      1	IN TXT		"abcdef" "ghijk"
OK
```

With this PR, nsd prints this as:
```
$ nsd-checkzone -p example example 
$ORIGIN .
example	1	IN	SOA	ns.example. admin.example. (
		1 7200 3600 1209600 3600 )
	1	IN	NS	example.
	1	IN	A	192.0.2.1
	1	IN	TXT	"ab" "cd" "ef\"gh\"" "ij" "kl"
$ORIGIN example.
test	1	IN	TXT	"abcdef" "ghijk"
; zone example is ok
```

So remaining issue is dotted_str before quoted-string (if we want to align that with bind too).

For reference, ldns-read zone says:
```
$ ldns-read-zone example 
example.	1	IN	SOA	ns.example. admin.example. 1 7200 3600 1209600 3600
example.	1	IN	NS	example.
example.	1	IN	A	192.0.2.1
example.	1	IN	TXT	"ab" "d" "ef\"gh\"" "ij" "kl\""
test.example.	1	IN	TXT	"abcdef" "ghijk"
```

### DIFF
--- a/dbcreate.c
+++ b/dbcreate.c
@@ -180,7 +180,7 @@ write_zone_to_udb(udb_base* udb, zone_type* zone, struct timespec* mtime,
 	return 1;
 }
 
-static int
+int
 print_rrs(FILE* out, struct zone* zone)
 {
 	rrset_type *rrset;
--- a/namedb.h
+++ b/namedb.h
@@ -369,6 +369,7 @@ int udb_write_rr(struct udb_base* udb, struct udb_ptr* z, rr_type* rr);
 void udb_del_rr(struct udb_base* udb, struct udb_ptr* z, rr_type* rr);
 int write_zone_to_udb(struct udb_base* udb, zone_type* zone,
 	struct timespec* mtime, const char* file_str);
+int print_rrs(FILE* out, struct zone* zone);
 /** marshal rdata into buffer, must be MAX_RDLENGTH in size */
 size_t rr_marshal_rdata(rr_type* rr, uint8_t* rdata, size_t sz);
 /* dbaccess.c */
--- a/nsd-checkzone.c
+++ b/nsd-checkzone.c
@@ -31,13 +31,14 @@ struct nsd nsd;
 static void
 usage (void)
 {
-	fprintf(stderr, "Usage: nsd-checkzone <zone name> <zone file>\n");
+	fprintf(stderr, "Usage: nsd-checkzone [-p] <zone name> <zone file>\n");
+	fprintf(stderr, "\t-p\tprint the zone if the zone is ok\n");
 	fprintf(stderr, "Version %s. Report bugs to <%s>.\n",
 		PACKAGE_VERSION, PACKAGE_BUGREPORT);
 }
 
 static void
-check_zone(struct nsd* nsd, const char* name, const char* fname)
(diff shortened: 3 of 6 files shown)


wcawijngaards (approved):
Perhaps the remaining issue could be solved by making the unquoted_dotted_str end if a `"` is seen.  It should not happen in dotted strings, I think, so that lexer change could make this parser change also solve the efgh issue?

Code looks good to me, and I like the nsd-checkzone feature to print it.

wcawijngaards:
This change is what I am proposing, could this fix the remaining issue?

It disallows the `"` in the charstr for unquoted dotted str and that makes the dotted str end if a quote is seen.  Thus the next parse block is the quoted string.  The escaped quote `\"` is still allowed, to write quotes in strings.

```diff
diff --git a/zlexer.lex b/zlexer.lex
index 2c3a3137..99b82821 100644
--- a/zlexer.lex
+++ b/zlexer.lex
@@ -130,7 +130,7 @@ SPACE   [ \t]
 LETTER  [a-zA-Z]
 NEWLINE [\n\r]
 ZONESTR [^ \t\n\r();.\"\$]|\\.|\\\n
-CHARSTR [^ \t\n\r();.]|\\.|\\\n
+CHARSTR [^ \t\n\r();.\"]|\\.|\\\n
 QUOTE   \"
 DOLLAR  \$
 COMMENT ;
```

wcawijngaards:
I tested the change, it works and prints `TXT	"ab" "cd" "ef" "gh" "ij" "kl"`.

PR #167, opened 2021-03-25
SVCB & HTTPS support

wtoorop:


configure.ac
```diff
@@ -1317,6 +1317,7 @@ AH_BOTTOM([
 #else
 #define ATTR_PACKED
 #endif
+
```
wtoorop:
```suggestion
```
 Don't add whitespace, it adds the suggestion that configure.ac changed.

dns.c
```diff
@@ -316,9 +316,41 @@ static rrtype_descriptor_type rrtype_descriptors[(RRTYPE_DESCRIPTORS_LENGTH+1)]
...
+	  , RDATA_WF_SVCPARAM, RDATA_WF_SVCPARAM, RDATA_WF_SVCPARAM
+	  , RDATA_WF_SVCPARAM, RDATA_WF_SVCPARAM, RDATA_WF_SVCPARAM
+	  },
+	  { RDATA_ZF_SHORT, RDATA_ZF_DNAME, RDATA_ZF_SVCPARAM
+	  , RDATA_ZF_SVCPARAM, RDATA_ZF_SVCPARAM, RDATA_ZF_SVCPARAM
+	  , RDATA_ZF_SVCPARAM, RDATA_ZF_SVCPARAM, RDATA_ZF_SVCPARAM
+	  , RDATA_ZF_SVCPARAM, RDATA_ZF_SVCPARAM, RDATA_ZF_SVCPARAM
+	  , RDATA_ZF_SVCPARAM, RDATA_ZF_SVCPARAM, RDATA_ZF_SVCPARAM
+	  , RDATA_ZF_SVCPARAM, RDATA_ZF_SVCPARAM, RDATA_ZF_SVCPARAM
+	  } },
```
wtoorop, line 415: Current definitions restrict the records to a maximum of 16 SvcParams. This is not necessarily bad, we have a similar approach with APL and SPF, but we might want to take the approach of TXT which has multiple TXT elements in a single RDATA_WF & RDATA_ZF.

dns.h
```diff
@@ -165,6 +167,14 @@ typedef enum nsd_rc nsd_rc_type;
...
 #define PSEUDO_TYPE_DLV	RRTYPE_DESCRIPTORS_LENGTH
 
+#define SVCB_KEY_MANDATORY		0
+#define SVCB_KEY_ALPN			1
+#define SVCB_KEY_NO_DEFAULT_ALPN	2
+#define SVCB_KEY_PORT			3
+#define SVCB_KEY_IPV4HINT		4
+#define SVCB_KEY_ECHCONFIG		5
+#define SVCB_KEY_IPV6HINT		6
+
```
wtoorop, line 178:
```suggestion
#define SVCPARAMKEY_COUNT 7

```

Move up from below (to keep SVCB related defines together)

dns.h
```diff
@@ -182,6 +192,8 @@ typedef enum nsd_rc nsd_rc_type;
 
 #define NSEC3_HASH_LEN 20
 
+#define SVCPARAMKEY_COUNT 7
+
```
wtoorop:
```suggestion
```
Accept if moved up.

rdata.c
```diff
@@ -642,6 +647,180 @@ rdata_loc_to_string(buffer_type *ATTR_UNUSED(output),
...
+		return 1;
+	} else
+		return 0; /* wireformat error*/
+}
+static int
+rdata_svcparam_mandatory_to_string(buffer_type *output, uint16_t val_len,
+	uint16_t *data)
+{
+	assert(val_len > 0); /* Guaranteed by rdata_svcparam_to_string */
+
```
wtoorop, line 728:
```suggestion

	if (val_len % sizeof(uint16_t))
		return 0; /* wireformat error, val_len must be multiple of shorts */
```

Without this test, the loop below may not end.

rdata.c
```diff
@@ -642,6 +647,180 @@ rdata_loc_to_string(buffer_type *ATTR_UNUSED(output),
...
+	if (size < 4)
+		return 0;
+	svcparamkey = ntohs(data[0]);
+
+	buffer_print_svcparamkey(output, svcparamkey);
+	val_len = ntohs(data[1]);
+	if (size != val_len + 4)
+		return 0; /* wireformat error */
+	if (!val_len)
+		return 1;
```
wtoorop: This allows SvcParams without values that should have values. Like all keys below (except for no-default-alpn). I suppose that is okay if it is also accepted on parsing.

zonec.c
```diff
@@ -27,6 +27,7 @@
 #endif
 
 #include <netinet/in.h>
+// #include <arpa/inet.h>
```
wtoorop:
```suggestion
```

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+		break;
+	}
+	if (key_len > sizeof(buf) - 1)
+		zc_error_prev_line("Unknown SvcParamKey");
+	else {
+		memcpy(buf, key, key_len);
+		buf[key_len] = 0;
+		zc_error_prev_line("Unknown SvcParamKey: %s", buf);
+	}
+	return 65535;
```
wtoorop:
```suggestion
	/* Although the returned value might be used by the caller,
	 * the parser has erred, so the zone will not be loaded.
	 */
	return -1;
```

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+zparser_conv_svcbparam_ipv4hint_value(region_type *region, const char *val)
+{
+	uint16_t *r;
+	int i, count;
+	char ip_str[INET_ADDRSTRLEN];
+	char *next_ip_str;
+	uint32_t *ip_wire_dst;
+
+	for (i = 0, count = 1; val[i]; i++)
+		count += (val[i] == ',');
```
wtoorop:
```suggestion
	for (i = 0, count = 1; val[i]; i++) {
		if (val[i] == ',')
			count += 1;
	}
```
The other looks nice, but not sure if it is 100% portable

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+
```
wtoorop, line 861:
```suggestion
	/* count == number of comma's in val + 1 
	 * so actually the number of IPv4 addresses in val
	 */
```

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+			assert(count == 1);
+
+		} else if (next_ip_str - val >= (int)sizeof(ip_str))
+			break;
+
+		else {
+			memcpy(ip_str, val, next_ip_str - val);
+			ip_str[next_ip_str - val] = 0;
+			if (inet_pton(AF_INET, ip_str, ip_wire_dst) != 1) {
+				val = ip_str;
```
wtoorop:
```suggestion
				val = ip_str; /* to use in error reporting below */
```

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+zparser_conv_svcbparam_ipv6hint_value(region_type *region, const char *val)
+{
+	uint16_t *r;
+	int i, count;
+	char ip6_str[INET6_ADDRSTRLEN];
+	char *next_ip6_str;
+	uint8_t *ipv6_wire_dst;
+
+	for (i = 0, count = 1; val[i]; i++)
+		count += (val[i] == ',');
```
wtoorop:
```suggestion
	for (i = 0, count = 1; val[i]; i++) {
		if (val[i] == ',')
			count += 1;
	}
```

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+
+	r = alloc_rdata(region, 2 * sizeof(uint16_t) + IP6ADDRLEN * count);
```
wtoorop, line 920:
```suggestion
	/* count == number of comma's in val + 1 
	 * so actually the number of IPv6 addresses in val
	 */
	r = alloc_rdata(region, 2 * sizeof(uint16_t) + IP6ADDRLEN * count);
```

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+			assert(count == 1);
+
+		} else if (next_ip6_str - val >= (int)sizeof(ip6_str))
+			break;
+
+		else {
+			memcpy(ip6_str, val, next_ip6_str - val);
+			ip6_str[next_ip6_str - val] = 0;
+			if (inet_pton(AF_INET6, ip6_str, ipv6_wire_dst) != 1) {
+				val = ip6_str;
```
wtoorop:
```suggestion
				val = ip6_str; /* for error reporting below */
```

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+	for (i = 0, count = 1; val[i]; i++)
+		count += (val[i] == ',');
+
+	r = alloc_rdata(region, (2 + count) * sizeof(uint16_t));
+	r[1] = htons(SVCB_KEY_MANDATORY);
+	r[2] = htons(sizeof(uint16_t) * count);
+	key_dst = (void *)&r[3];
+
+	// @TODO add error for forbidden keys
+
```
wtoorop:
```suggestion
```
That is already done in zadd_rdata_svcb_check_wireformat() ?

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+{
+	uint16_t *r;
+	int key, prev_key;
+	int i, count;
+	char* next_key;
+	uint16_t* key_dst;
+
+	for (i = 0, count = 1; val[i]; i++)
+		count += (val[i] == ',');
+
```
wtoorop:
```suggestion
	for (i = 0, count = 1; val[i]; i++) {
		if (val[i] == ',')
			count += 1;
	}

```

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+						   "SvcParam\n");
+				break;
+			case SVCB_KEY_ALPN:
+				zc_error_prev_line("alpn is automatically "
+				                   "mandatory and may not be "
+						   "used in a mandatory "
+						   "SvcParam\n");
+
+				break;
+		}
```
wtoorop:
```suggestion
```
Did we not find that current_rr.type is not yet set at this point?
Are these checks not done in zadd_rdata_svcb_check_wireformat()?

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+	/* In draft-ietf-dnsop-svcb-https-04 Section 7:
+	 *
+	 *     Keys (...) MUST NOT appear more than once.
+	 */
+	prev_key = -1;
+	for(i = 0; i < count; i++) {
+		key = read_uint16((void *)&r[3 + i]);
+
+		if (key == SVCB_KEY_MANDATORY)
+			zc_error_prev_line("mandatory should not be a value of the mandatory key\n");
```
wtoorop: Is this not also done in zadd_rdata_svcb_check_wireformat()?
Perhaps the whole scan for doubles should be done there instead of here, to not loop over the mandatory keys twice?

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+		                  , key, val);
+	}
+	/* assert(val == NULL); */
+	if ((eq = memchr(key, '=', key_len))) {
+		size_t new_key_len = eq - key;
+
+		return zparser_conv_svcbparam_key_value(
+		    region, key, new_key_len, eq+1, key_len - new_key_len - 1);
+	}
+	/* SvcParam is only a SvcParamKey */
```
wtoorop, line 1188: So this looks like indeed a echconfig (and other SvcParams that have values) without value is possible... maybe this is just okay... WDYT @TCY16 ?

zonec.c
```diff
@@ -1215,6 +1599,116 @@ zadd_rdata_txt_clean_wireformat()
...
+{
+	return ((int)read_uint16(rdata_atom_data(*(rdata_atom_type *)a)))
+	     - ((int)read_uint16(rdata_atom_data(*(rdata_atom_type *)b)));
+}
+
+void
+zadd_rdata_svcb_check_wireformat(uint16_t rr_type)
+{
+	size_t i, j;
+	uint8_t paramkeys[SVCPARAMKEY_COUNT - 1];
```
wtoorop: -1 Stack violation with ipv6hints, which probably did not result in crash because of aligning on 32bits or 64bits.
```suggestion
	uint8_t paramkeys[65536];
```

zonec.c
```diff
@@ -1215,6 +1599,116 @@ zadd_rdata_txt_clean_wireformat()
...
+	qsort( (void *)&parser->current_rr.rdatas[2]
+	     , parser->current_rr.rdata_count - 2
+	     , sizeof(rdata_atom_type *)
+	     , svcparam_key_cmp
+	     );
+
+	for (i = 2; i < parser->current_rr.rdata_count; i++) {
+		// @TODO remove after debugging is done
+		// size_t    size = rdata_atom_size(parser->current_rr.rdatas[i]);
+
```
wtoorop:
```suggestion
```
cleanup ;)

zonec.c
```diff
@@ -1215,6 +1599,116 @@ zadd_rdata_txt_clean_wireformat()
...
+		uint8_t  *data = rdata_atom_data(parser->current_rr.rdatas[i]);
+		key  = read_uint16(data);
+
+		/* keep track of keys that are present*/
+		if (key < SVCPARAMKEY_COUNT) {
+			paramkeys[key] = 1;
+		}
+
+		// @TODO check doubles
```
wtoorop: Yes this is a TODO... is there text in the draft?

zonec.c
```diff
@@ -1215,6 +1599,116 @@ zadd_rdata_txt_clean_wireformat()
...
+	 * and "no-default-alpn". 
+	 * (...) 
+	 * "automatically mandatory", i.e. mandatory if they arepresent in an RR.
+	 */
+	if (rr_type == TYPE_HTTPS) {
+		if (!(paramkeys[SVCB_KEY_PORT]) || !(paramkeys[SVCB_KEY_NO_DEFAULT_ALPN])) {
+			zc_error_prev_line("port and no-default-alpn must be included"
+			                   " in the HTTPS record");
+		}
+	}
```
wtoorop:
```suggestion
	if (rr_type == TYPE_HTTPS
	&& (!paramkeys[SVCB_KEY_PORT] || !paramkeys[SVCB_KEY_NO_DEFAULT_ALPN])) {
		zc_error_prev_line("port and no-default-alpn must be included"
			                   " in the HTTPS record");
	}
```
I think leaving out unnecessary nesting and brackets increase readability, but feel free to disagree.

zonec.c
```diff
@@ -1215,6 +1599,116 @@ zadd_rdata_txt_clean_wireformat()
...
+
+	/* Check that the mandatory key is present */
+	if (paramkeys[SVCB_KEY_MANDATORY]) {
+		size_t    size = rdata_atom_size(parser->current_rr.rdatas[2]);
+		uint16_t  *mandatory_values = (void*)rdata_atom_data(parser->current_rr.rdatas[2]);
+		key  = read_uint16(mandatory_values);
+
+		mandatory_values += 2; /* skip the key type and length */
+
+		for (i = 0; i < (size - 4)/2; i++) {
```
wtoorop: Maybe add test for odd mandatory rdata?

wtoorop:
```suggestion
		if (size % 2)
			zc_error_prev_line("mandatory rdata must be a even");
			
		else for (i = 0; i < (size - 4)/2; i++) {
```
I know this is impossible when parsing, but it is scary to have a potential never ending loop without this check.
You never know how this will be called in the future... (from unknown rdata format for example).

zonec.c
```diff
@@ -1215,6 +1599,116 @@ zadd_rdata_txt_clean_wireformat()
...
+			                   " in the HTTPS record");
+		}
+	}
+
+	/* Check that the mandatory key is present */
+	if (paramkeys[SVCB_KEY_MANDATORY]) {
+		size_t    size = rdata_atom_size(parser->current_rr.rdatas[2]);
+		uint16_t  *mandatory_values = (void*)rdata_atom_data(parser->current_rr.rdatas[2]);
+		key  = read_uint16(mandatory_values);
+
```
wtoorop:
```suggestion
```

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+	r = alloc_rdata(region, (2 + count) * sizeof(uint16_t));
+	r[1] = htons(SVCB_KEY_MANDATORY);
+	r[2] = htons(sizeof(uint16_t) * count);
+	key_dst = (void *)&r[3];
+
+	// @TODO add error for forbidden keys
+
+	for(;;) {
+		if (!(next_key = strchr(val, ','))) {
+			*key_dst = svcbparam_lookup_key(val, val_len);
```
wtoorop:
```suggestion
			*key_dst = htons(svcbparam_lookup_key(val, val_len));
```

zonec.c
```diff
@@ -753,6 +754,389 @@ zparser_conv_nsec(region_type *region,
...
+			break;	
+		} else {
+			*key_dst = svcbparam_lookup_key(val, next_key - val);
```
wtoorop:
```suggestion
			*key_dst = htons(svcbparam_lookup_key(val, next_key - val));
```

rdata.c
```diff
@@ -642,6 +647,180 @@ rdata_loc_to_string(buffer_type *ATTR_UNUSED(output),
...
+		return 0; /* wireformat error*/
+}
+static int
+rdata_svcparam_mandatory_to_string(buffer_type *output, uint16_t val_len,
+	uint16_t *data)
+{
+	assert(val_len > 0); /* Guaranteed by rdata_svcparam_to_string */
+
+	buffer_write_u8(output, '=');
+	buffer_print_svcparamkey(output, *data);
```
wtoorop:
```suggestion
	buffer_print_svcparamkey(output, ntohs(*data));
```

rdata.c
```diff
@@ -642,6 +647,180 @@ rdata_loc_to_string(buffer_type *ATTR_UNUSED(output),
...
+	data += 1;
+
+	while ((val_len -= sizeof(uint16_t))) {
+		buffer_write_u8(output, ',');
+		buffer_print_svcparamkey(output, *data);
```
wtoorop:
```suggestion
		buffer_print_svcparamkey(output, ntohs(*data));
```

zonec.c
```diff
@@ -1215,6 +1599,116 @@ zadd_rdata_txt_clean_wireformat()
...
+	/* Check that the mandatory key is present */
+	if (paramkeys[SVCB_KEY_MANDATORY]) {
+		size_t    size = rdata_atom_size(parser->current_rr.rdatas[2]);
+		uint16_t  *mandatory_values = (void*)rdata_atom_data(parser->current_rr.rdatas[2]);
+		key  = read_uint16(mandatory_values);
+
+		mandatory_values += 2; /* skip the key type and length */
+
+		for (i = 0; i < (size - 4)/2; i++) {
+			key = mandatory_values[i];
```
wtoorop:
```suggestion
			key = ntohs(mandatory_values[i]);
```

zonec.c
```diff
@@ -1215,6 +1599,116 @@ zadd_rdata_txt_clean_wireformat()
...
+		// @TODO remove after debugging is done
+		// size_t    size = rdata_atom_size(parser->current_rr.rdatas[i]);
+
+		uint8_t  *data = rdata_atom_data(parser->current_rr.rdatas[i]);
+		key  = read_uint16(data);
+
+		/* keep track of keys that are present*/
+		if (key < SVCPARAMKEY_COUNT) {
+			paramkeys[key] = 1;
+		}
```
wtoorop:
```suggestion
		paramkeys[key] = 1;
```

zonec.c
```diff
@@ -1215,6 +1599,116 @@ zadd_rdata_txt_clean_wireformat()
...
+			if (key < SVCPARAMKEY_COUNT && !(paramkeys[key]))
+				zc_error_prev_line("mandatorySvcParamKey: %s is missing "
+				                   "the record\n", svcparamkey_strs[key]);
+
+			/* In draft-ietf-dnsop-svcb-https-04 Section 8
+			 * automatically mandatory MUST NOT appear in its own value-list
+			 */
+			else if (key < SVCPARAMKEY_COUNT && paramkeys[key]
+				 && (rr_type == TYPE_HTTPS)
+				 && (key == SVCB_KEY_PORT || key == SVCB_KEY_NO_DEFAULT_ALPN))
```
wtoorop:
```suggestion
			else if (rr_type == TYPE_HTTPS
				 && (key == SVCB_KEY_PORT || key == SVCB_KEY_NO_DEFAULT_ALPN))
```

zonec.c
```diff
@@ -1215,6 +1599,116 @@ zadd_rdata_txt_clean_wireformat()
...
+					if (key == read_uint16(temp_data)) {
+						found = 1;
+						break;
+					}
+				}
+
+				if (!(found))
+					zc_error_prev_line("mandatorySvcParamKey: key%hu is missing "
+					                   "the record\n", key);
+			}
```
wtoorop:
```suggestion
```

zonec.c
```diff
@@ -1215,6 +1599,116 @@ zadd_rdata_txt_clean_wireformat()
...
+
+		mandatory_values += 2; /* skip the key type and length */
+
+		for (i = 0; i < (size - 4)/2; i++) {
+			key = mandatory_values[i];
+
+			if (key < SVCPARAMKEY_COUNT && !(paramkeys[key]))
+				zc_error_prev_line("mandatorySvcParamKey: %s is missing "
+				                   "the record\n", svcparamkey_strs[key]);
+
```
wtoorop:
```suggestion
			else if (!paramkeys[key])
				zc_error_prev_line("mandatorySvcParamKey: key%hu is missing "
				                   "the record\n", key);
```

zonec.c
TCY16: Implemented

zonec.c
TCY16: added (by Willem) and approved

zonec.c
TCY16: resolved

dns.c
TCY16: I agree, this should be reworked

TCY16: resolved in latest commit

rdata.c
TCY16: resolved in latest commit

zonec.c
TCY16: resolved in latest commit

wtoorop:
Still TODO:
- [x] Test package
- [x] Errors turn into warnings when zones are loaded in the secondary role

rdata.c
```diff
@@ -642,6 +647,210 @@ rdata_loc_to_string(buffer_type *ATTR_UNUSED(output),
...
+static int
+rdata_svcparam_ipv4hint_to_string(buffer_type *output, uint16_t val_len,
+	uint16_t *data)
+{
+	char ip_str[INET_ADDRSTRLEN];
+	
+	assert(val_len > 0); /* Guaranteed by rdata_svcparam_to_string */
+
+	if ((val_len % IP4ADDRLEN) == 0) {
+		buffer_printf(output, "=%s", inet_ntop(AF_INET, data, ip_str, sizeof(ip_str)));
```
wcawijngaards: The function inet_ntop can return NULL, you have to check the return value from it. Also in line 682.

rdata.c
```diff
@@ -642,6 +647,210 @@ rdata_loc_to_string(buffer_type *ATTR_UNUSED(output),
...
+		return 0; /* wireformat error, a short is 2 bytes */
+	buffer_printf(output, "=%d", (int)ntohs(data[0]));
+	return 1;
+}
+
+static int
+rdata_svcparam_ipv4hint_to_string(buffer_type *output, uint16_t val_len,
+	uint16_t *data)
+{
+	char ip_str[INET_ADDRSTRLEN];
```
wcawijngaards: Does this need a +1 on the array size for terminating zero?

rdata.c
```diff
@@ -642,6 +647,210 @@ rdata_loc_to_string(buffer_type *ATTR_UNUSED(output),
...
+		buffer_printf(output, "=%s", inet_ntop(AF_INET, data, ip_str, sizeof(ip_str)));
+		data += IP4ADDRLEN / sizeof(uint16_t);
+
+		while ((val_len -= IP4ADDRLEN)) {
+			buffer_printf(output, ",%s", inet_ntop(AF_INET, data, ip_str, sizeof(ip_str)));
+			data += IP4ADDRLEN / sizeof(uint16_t);
+		}
+		return 1;
+	} else
+		return 0; /* wireformat error, a short is 2 bytes */
```
wcawijngaards: This comment is wrong, it is not about the length of a short, here.

rdata.c
```diff
@@ -642,6 +647,210 @@ rdata_loc_to_string(buffer_type *ATTR_UNUSED(output),
...
+{
+	char ip_str[INET_ADDRSTRLEN];
+	
+	assert(val_len > 0); /* Guaranteed by rdata_svcparam_to_string */
+
+	if ((val_len % IP4ADDRLEN) == 0) {
+		buffer_printf(output, "=%s", inet_ntop(AF_INET, data, ip_str, sizeof(ip_str)));
+		data += IP4ADDRLEN / sizeof(uint16_t);
+
+		while ((val_len -= IP4ADDRLEN)) {
```
wcawijngaards: I would appreciate the explicit check for nonzero, because that makes it more readable with the `-=` side effect.

```suggestion
		while ((val_len -= IP4ADDRLEN) > 0) {
```

rdata.c
```diff
@@ -642,6 +647,210 @@ rdata_loc_to_string(buffer_type *ATTR_UNUSED(output),
...
+		return 1;
+	} else
+		return 0; /* wireformat error, a short is 2 bytes */
+}
+
+static int
+rdata_svcparam_ipv6hint_to_string(buffer_type *output, uint16_t val_len,
+	uint16_t *data)
+{
+	char ip_str[INET6_ADDRSTRLEN];
```
wcawijngaards: Does this need +1 on the array size for terminating zero?

rdata.c
```diff
@@ -642,6 +647,210 @@ rdata_loc_to_string(buffer_type *ATTR_UNUSED(output),
...
+	
+	assert(val_len > 0); /* Guaranteed by rdata_svcparam_to_string */
+
+	if ((val_len % IP6ADDRLEN) == 0) {
+		buffer_printf(output, "=%s", inet_ntop(AF_INET6, data, ip_str, sizeof(ip_str)));
+		data += IP6ADDRLEN / sizeof(uint16_t);
```
wcawijngaards, line 709: similar comments to the ip4 case.

rdata.c
wcawijngaards: Yes it does, Unbound has, in similar code snippets, a +1 for that array size.

rdata.c
wcawijngaards: Yes, it seems so.

zonec.c
```diff
@@ -753,6 +753,426 @@ zparser_conv_nsec(region_type *region,
...
+	zc_error_prev_line("Could not parse port SvcParamValue: \"%s\"", val);
+	return alloc_rdata_init(region, "", 0);
+}
+
+static uint16_t *
+zparser_conv_svcbparam_ipv4hint_value(region_type *region, const char *val)
+{
+	uint16_t *r;
+	int count;
+	char ip_str[INET_ADDRSTRLEN];
```
wcawijngaards: Add one to it.
```suggestion
	char ip_str[INET_ADDRSTRLEN+1];
```

zonec.c
```diff
@@ -753,6 +753,426 @@ zparser_conv_nsec(region_type *region,
...
+
+	return r;
+}
+
+static uint16_t *
+zparser_conv_svcbparam_ipv6hint_value(region_type *region, const char *val)
+{
+	uint16_t *r;
+	int i, count;
+	char ip6_str[INET6_ADDRSTRLEN];
```
wcawijngaards: Add one for a terminating byte.
```suggestion
	char ip6_str[INET6_ADDRSTRLEN+1];
```

zonec.c
```diff
@@ -753,6 +753,426 @@ zparser_conv_nsec(region_type *region,
...
+			count += 1;
+	}
+
+	/* count == number of comma's in val + 1 
+	 * so actually the number of IPv4 addresses in val
+	 */
+	r = alloc_rdata(region, 2 * sizeof(uint16_t) + IP4ADDRLEN * count);
+	r[1] = htons(SVCB_KEY_IPV4HINT);
+	r[2] = htons(IP4ADDRLEN * count);
+	ip_wire_dst = (void *)&r[3];
```
wcawijngaards, line 868: Should limit the max of count, somewhere before the alloc_rdata call. It must be smaller than (max uint16_t) / IP4ADDRLEN. But is also limited by rdatasize and packetsize. The 16k max that that gives is not even that big, and could conceivable be on some file input line, even more so for IPv6.

zonec.c
```diff
@@ -753,6 +753,426 @@ zparser_conv_nsec(region_type *region,
...
+				break;
+
+			assert(count == 1);
+
+		} else if (next_ip_str - val >= (int)sizeof(ip_str))
+			break;
+
+		else {
+			memcpy(ip_str, val, next_ip_str - val);
+			ip_str[next_ip_str - val] = 0;
```
wcawijngaards, line 882: This is a buffer overflow if next_ip_str - val == (int)sizeof(ip_str), right? Then that is equal to the size of the array, and this sets ip_str[one too far] = 0; and that can be solved by checking a bit more earlier.

wcawijngaards: Oh wait no, that is exactly covered by the check if statement above, so should not happen. Sorry.

tpkg/svcb.tdir/svcb.pre
```diff
@@ -0,0 +1,30 @@
+# #-- svcb.pre--#
+# source the master var file when it's there
+[ -f ../.tpkg.var.master ] && source ../.tpkg.var.master
+# use .tpkg.var.test for in test variable passing
+[ -f .tpkg.var.test ] && source .tpkg.var.test
+. ../common.sh
+
+# start NSD
+get_random_port 2
```
wcawijngaards: Because the test uses three ports.

```suggestion
get_random_port 3
```

tpkg/svcb.tdir/svcb.pre
```diff
@@ -0,0 +1,30 @@
...
+echo "export TPKG_NSD_PRI_PID=$TPKG_NSD_PRI_PID" >> .tpkg.var.test
+echo "export TPKG_SEC_PORT=$TPKG_SEC_PORT" >> .tpkg.var.test
+echo "export TPKG_NSD_SEC_PID=$TPKG_NSD_SEC_PID" >> .tpkg.var.test
+
+sed -e "s/@TPKG_PRI_PORT@/$TPKG_PRI_PORT/g" \
+    -e "s/@TPKG_CTRL_PORT@/$TPKG_CTRL_PORT/g" svcb.secondary.conf.in > svcb.secondary.conf
+
+$TPKG_NSD -c svcb.primary.conf -u $LOGNAME -p $TPKG_PRI_PORT -P $TPKG_NSD_PRI_PID 
+$TPKG_NSD -c svcb.secondary.conf -u $LOGNAME -p $TPKG_SEC_PORT -P $TPKG_NSD_SEC_PID 
+wait_nsd_up nsd.log
```
wcawijngaards, line 30: And also `wait_nsd_up nsd-secondary.log`, here?
```suggestion
wait_nsd_up nsd.log
wait_nsd_up nsd-secondary.log
```

zonec.c
```diff
@@ -753,6 +753,426 @@ zparser_conv_nsec(region_type *region,
...
+{
+	uint16_t *r;
+	size_t i, count;
+	char* next_key;
+	uint16_t* key_dst;
+
+	for (i = 0, count = 1; val[i]; i++) {
+		if (val[i] == ',')
+			count += 1;
+	}
```
wcawijngaards, line 976: It is a good idea to put a max on the count here and fail if too large. Stops too large memory allocations, and 2*count > maxuint16_t.

zonec.c
```diff
@@ -1215,6 +1635,121 @@ zadd_rdata_txt_clean_wireformat()
...
+			if(zone_is_slave(parser->current_zone->opts))
+				zc_warning_prev_line("mandatory MUST not be included"
+						     " as mandatory parameter");
+			else
+				zc_error_prev_line("mandatory MUST not be included"
+						   " as mandatory parameter");
+		}
+		if (key == prev_key) {
+			if(zone_is_slave(parser->current_zone->opts))
+				zc_warning_prev_line("Keys in SVcParam mandatory "
```
wcawijngaards: You probably mean s/SVc/Svc/

wcawijngaards (approved):
Code looks good!

zonec.c
```diff
@@ -1215,6 +1647,121 @@ zadd_rdata_txt_clean_wireformat()
...
+			if(zone_is_slave(parser->current_zone->opts))
+				zc_warning_prev_line("mandatory MUST not be included"
+						     " as mandatory parameter");
+			else
+				zc_error_prev_line("mandatory MUST not be included"
+						   " as mandatory parameter");
+		}
+		if (key == prev_key) {
+			if(zone_is_slave(parser->current_zone->opts))
+				zc_warning_prev_line("Keys inSvcParam mandatory "
```
wcawijngaards, line 1788: Fix in your, otherwise very nice, list of fix commits.

```suggestion
				zc_warning_prev_line("Keys in SvcParam mandatory "
```

Issue #168, opened 2021-04-02
Buffer overflow in the dname_to_string() function


bsdb0y:
Hi, 

While fuzzing nsd-checkzone in NSD 4.2.4 (and git nighly build (revision: a1879fb4363cb0ad1fa85ef9a11499c7e3d95540)), I found a buffer overflow in the dname_to_string() function, in dname.c.

Attaching a reproducer (zipped so GitHub accepts it): [input_test0.zip](https://github.com/NLnetLabs/nsd/files/6250947/input_test0.zip)

Issue can be reproduced by running:

`nsd-checkzone all.rr input_test0`

```
=================================================================
==2301135==ERROR: AddressSanitizer: global-buffer-overflow on address 0x000000f7e91b at pc 0x00000043bb11 bp 0x7ffe3f9c8c20 sp 0x7ffe3f9c83b8
WRITE of size 5 at 0x000000f7e91b thread T0
    #0 0x43bb10 in vsnprintf (/src/nsd-NSD_4_2_4_REL/nsd-checkzone+0x43bb10)
    #1 0x43d060 in snprintf (/src/nsd-NSD_4_2_4_REL/nsd-checkzone+0x43d060)
    #2 0x4e421f in dname_to_string /src/nsd-NSD_4_2_4_REL/dname.c:423:5
    #3 0x61ff60 in domain_to_string /src/nsd-NSD_4_2_4_REL/./namedb.h:315:10
    #4 0x61e335 in process_rr /src/nsd-NSD_4_2_4_REL/zonec.c:1435:79
    #5 0x623f0f in yyparse /src/nsd-NSD_4_2_4_REL/./zparser.y:125:8
    #6 0x620ce6 in zonec_read /src/nsd-NSD_4_2_4_REL/zonec.c:1627:2
    #7 0x63ffff in check_zone /src/nsd-NSD_4_2_4_REL/nsd-checkzone.c:61:11
    #8 0x63fc79 in main /src/nsd-NSD_4_2_4_REL/nsd-checkzone.c:131:2
    #9 0x7feb592d60b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
    #10 0x41da0d in _start (/src/nsd-NSD_4_2_4_REL/nsd-checkzone+0x41da0d)

0x000000f7e91b is located 0 bytes to the right of global variable 'buf' defined in 'dname.c:391:14' (0xf7e420) of size 1275
[...]

PR #169, opened 2021-04-05
Do not try to print error_domain or error_dname when printing errors.

wtoorop:
Resolves issue #168

### DIFF
--- a/zonec.c
+++ b/zonec.c
@@ -1429,6 +1429,11 @@ process_rr(void)
 		rr->owner->is_apex = 1;
 	}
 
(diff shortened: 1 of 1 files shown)


wcawijngaards:
The code should move to around line 1412, before the zone soa check. Because that accesses the rr->owner element, that could be broken. The has_soa and is_apex lines should not be happening on an error_domain, I should think, so move the check earlier to stop that from potentially happening.

TCY16 (approved):

wcawijngaards (approved):
Looks good, I do not see other issues.

PR #173, opened 2021-05-04
Bugfix/dnstap out of sync

wtoorop:
Recovery of dnstap collector getting out of sync with the serve childs.
TODO:

- [x] respawn dnstap collector on reload without closing sockets like the serve childs (otherwise we leak memory on zone updates because the dnstap collector has old version of database...)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -7,45 +7,32 @@
 # For now builds are limited to Linux, FreeBSD and macOS builds on AMD64, more
 # may be added in the future.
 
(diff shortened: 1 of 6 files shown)


wcawijngaards (approved):
Looks good, respawn every reload keeps track of the zones in memory. Could also free that data in the new dnstap-collector process, but this may work just as well and could be simpler.

wcawijngaards (approved):
Looks good to me.

wtoorop:
Closing in favor of #182

PR #182, opened 2021-06-29
dnstap out of sync

wtoorop:
This has fixes that alternates the pipes used to write to the dnstap-collector on every reload, in the hope to prevent clashes (writing on the same pipe simultaneously). This caused NSD's dnstap-collector to get into an undefined state. This PR has also code to stop and restart the dnstap-collector in such situations to recover from it.


wtoorop:
TODO:
- [ ] Restarting of dnstap collector if it exited during reload

wcawijngaards (approved):
datagram socket looks like it could solve the problems. Perhaps the socket swap is not needed with it, but should not be a problem.

PR #183, opened 2021-07-06
Shivan Kaul XoT review

wtoorop:
These are nits that came out of reviewing client side XoT from @ShivanKaul and @saradickinson .
I am planning to create a tpkg test for it. For that it is convenient to have a `tls-cert-bundle` like how that exists in Unbound, so I'll create an option for that first.


xfrd-tcp.c
```diff
@@ -573,12 +701,62 @@ xfrd_tcp_open(struct xfrd_tcp_set* set, struct xfrd_tcp_pipeline* tp,
...
+		if(!ssl_handshake(tp)) {
+			if(tp->handshake_want == SSL_ERROR_SYSCALL) {
+				log_msg(LOG_ERR, "xfrd: TLS handshake failed "
+					"for %s to %s: %s", zone->apex_str,
+					zone->master->ip_address_spec,
+					strerror(errno));
+
+			} else if(tp->handshake_want == SSL_ERROR_SSL) {
+				if(SSL_get_verify_result(tp->ssl) == X509_V_OK)
+					log_crypto_err("xfrd: TLS handshake failed");
```
wcawijngaards:
```suggestion
			} else if(tp->handshake_want == SSL_ERROR_SSL) {
				log_crypto_err("xfrd: TLS handshake failed");
```
The get_verify_result is only necessary when the handshake has completed successfully? With SSL_ERROR_SSL we know it is wrong and the log_crypto_err function prints the details that are in the error stack.

xfrd-tcp.c
```diff
@@ -737,7 +988,33 @@ xfrd_tcp_write(struct xfrd_tcp_pipeline* tp, xfrd_zone_type* zone)
...
+			return;
+
+		} else {
+			if(tp->handshake_want == SSL_ERROR_SYSCALL) {
+				log_msg(LOG_ERR, "xfrd: TLS handshake failed: %s",
+					strerror(errno));
+
+			} else if(tp->handshake_want == SSL_ERROR_SSL) {
+				if (SSL_get_verify_result(tp->ssl) == X509_V_OK)
+					log_crypto_err("xfrd: TLS handshake failed");
```
wcawijngaards:
```suggestion
			} else if(tp->handshake_want == SSL_ERROR_SSL) {
				log_crypto_err("xfrd: TLS handshake failed");
```
Same as the other get_verify_result comment?

xfrd-tcp.c
wtoorop: The `tls_verify_callback()` function also prints an error message if the name doesn't match.
I thought to be smart and prevent double printing of the message by first testing if something else occurred  than a verification problem -- since `SSL_get_verify_result()` results `X509_V_OK` if no verification has taken place yet.
But I noticed with the xot.tdir test that no error is logged when the CA doesn't match. So I guess it is better to have double error reporting than no reporting at all in some error conditions.
Will remove the extra `SSL_get_verify_result(tp->ssl) == X509_V_OK` test

xfrd-tcp.c
wtoorop: Yes I'll remove that extra test...

xfrd-tcp.c
wtoorop: Ah, it was actually a type in the xot test that caused the missing CA not to be printed, but I think it is fine (and safe) to just report the error twise (once during verification, and once when the handshake fails). The fixes are in commits 78eace0 and 9603146

wcawijngaards (approved):
The code looks good to merge. The nsd-control ports in the tdir test are unused.

PR #185, opened 2021-07-28
Mutual TLS

cesarkuroiwa:
Add new config parameters to the `tls-auth` section for client authentication by the server in a XFR-over-TLS operation (Mutual TLS)


wcawijngaards (approved):
Changes look good. Maybe we can add this to the code repository.

wcawijngaards:
Thank you for the feature! It allows NSD to use client authentication for AXFR and IXFR downloads over TLS. I have merged it and made a couple of small fixes to make it nicer.

PR #187, opened 2021-08-11
- Support using system-wide crypto policies

paulwouters:
System Wide crypto policies are supported via crypto libraries
(openssl, NSS, gnutls) on various distros (RHEL, CentOS, Fedora)

By not manually calling SSL_CTX_set_cipher_list(), the system-wide
crypto policies would be used instead.

System-wide crypto policies are detected via the presence of the
file /etc/crypto-policies/config

### DIFF
--- a/server.c
+++ b/server.c
@@ -1995,9 +1995,12 @@ server_tls_ctx_setup(char* key, char* pem, char* verifypem)
 	}
 #endif
 #if defined(SHA256_DIGEST_LENGTH) && defined(SSL_TXT_CHACHA20)
-	/* if we have sha256, set the cipher list to have no known vulns */
-	if(!SSL_CTX_set_cipher_list(ctx, "ECDHE+AESGCM:ECDHE+CHACHA20"))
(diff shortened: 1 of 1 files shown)


paulwouters:
note that for example, when the system is running in FIPS mode, the crypto policies would not allow CHACHA20 which is currently being set by nsd. By using the system-wide crypto policies when available, these issues go away.

wcawijngaards (approved):
Looks good to automatically detect what the correct settings are.

wcawijngaards:
Thank you for the fix. The detect is very nice, and I also ported it to unbound for a fix to its cipher list. That cipher list was not as restrictive as this one, so the fix is nice anyway. Unbound also allows the cipher list to be set in config, for the public facing endpoints.

paulwouters:
For unbound, I had fixed it differently, by just setting it directly in the unbound.conf file we ship.

But see this discussion: https://gitlab.com/redhat-crypto/fedora-crypto-policies/-/issues/32

PR #192, opened 2021-09-01
Features/credns

wtoorop:
Fresh merge & review of Jeroen's zone verification branch


k0ekk0ek:
Rebased on PR #209 by @wcawijngaards. Ran the tests locally, fixed a couple of them, but they all pass now. We probably want to refresh this once the PR this is based on gets merged.

k0ekk0ek:
This PR is now rebased on the current master branch. Locally at least the added tests pass. The unit tests and the Clang analysis test does as well. I'd like extra sets of eyes to go over this. Btw, I bumped the minor version as it's a new feature etc.

k0ekk0ek:
@wtoorop, I cannot select you as a reviewer (probably because you created the PR in the first place?), but if you would be so kind.

PR #193, opened 2021-09-10
Xfrd tcp max memory

wcawijngaards:
This adds configuration options for the zone transfers memory usage. It sets the maximum number of transfers that can be performed, at the same time.

The options added are `xfrd-tcp-max` and `xfrd-tcp-pipeline`.

### DIFF
--- a/xfrd.h
+++ b/xfrd.h
@@ -240,7 +240,7 @@ enum xfrd_packet_result {
    Note that also some sockets are used for writing the ixfr.db, xfrd.state
    files and for the pipes to the main parent process.
 
-   For xfrd_tcp_max 128 can be the number of TCP AXFR/IXFR concurrent
+   For xfrd_tcp_max, 128 is the default number of TCP AXFR/IXFR concurrent
    connections. Each entry has 64Kb buffer preallocated.
 */
 #define XFRD_MAX_UDP 128 /* max number of UDP sockets at a time for IXFR */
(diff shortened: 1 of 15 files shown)


TCY16 (approved):

xfrd.h
```diff
@@ -239,9 +239,10 @@ enum xfrd_packet_result {
    And it should be below FD_SETSIZE, to be able to select() on replies.
    Note that also some sockets are used for writing the ixfr.db, xfrd.state
    files and for the pipes to the main parent process.
+
+   For xfrd_tcp_max 128 can be the number of TCP AXFR/IXFR concurrent
```
wtoorop: This fixed setting changed to a configurable setting, but the default value changed as well.
```suggestion
   For xfrd_tcp_max , 32 is the default number of TCP AXFR/IXFR concurrent
```

wtoorop (approved):
Thank you @wcawijngaards !
I reviewed now three times thoroughly and it all looks very good.
One question. We're changing the default value of maximum number of TCP connections to do transfers from NSD from 128 to 32. Do you think that's okay? We're usually quite conservative about those things. Even if we would keep xfrd_tcp_max to be 128 there will be much less memory usage because of replacing the 64k arrays of IDs and zone pointers with xfrd_tcp_pipeline number of ids and zone pointers. That's also a change of defaults, but I suspect this will be much less noticable (if at all) in operations.
Also, sorry this took me so long.

xfrd.h
wcawijngaards: Fixed this in https://github.com/NLnetLabs/nsd/pull/193/commits/ed8117cc2c067fb5b4a6db08e819503c3842f7bf
So I used 32 because the 64K preallocated buffer is large, and with 32 that is a much smaller footprint than with 128. But it is still not that big with 128, so I reverted that back to a default of 128. So there is no change of default in the PR.

PR #207, opened 2022-02-05
Sync nsd-control-setup with unbound-control-setup to generate certificates with SANs

sbiberhofer:
Right now `nsd-control-setup` generates very simple certificates which do not contain a SAN attribute. This is particularly problematic for tools like `nsd_exporter`, which requires SANs when compiled with recent versions of Go. 

Since this has already been fixed in unbound's version of this script, this PR basically syncs `nsd-control-setup` with `unbound-control-setup` to fix this issue in what I hope is the simplest possible way. Aside from adding the necessary extensions, this also extends certificate lifetime.

### DIFF
--- a/nsd-control-setup.sh.in
+++ b/nsd-control-setup.sh.in
@@ -43,7 +43,7 @@ SERVERNAME=nsd
 CLIENTNAME=nsd-control
 
 # validity period for certificates
-DAYS=3650
+DAYS=7200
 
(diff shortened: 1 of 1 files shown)


wcawijngaards (approved):
These changes look good and match unbound's version.

wcawijngaards:
Thanks for the patch! It is a good idea to also have those fixes here.

PR #209, opened 2022-03-22
IXFR out

wcawijngaards:
IXFR support to provide IXFR going out, towards requestors. The implementation can store downloaded IXFRs, with `store-ixfr: yes` and create IXFRs from zonefiles with `create-ixfr: yes`.

The `ixfr-number:5` and `ixfr-size: 1048576` settings specify the maximum amount to store for IXFR data.

IXFRs are stored next to the zonefile in '.ixfr' files.

NSD already supports downloading IXFRs and using them, this change also makes it possible to provide IXFRs to requestors.


k0ekk0ek:
Agree on the mechanics. Didn't try to spot programming errors yet since this is still WiP and testcases etc are still coming in. If you want I'll do so at a later stage.

ixfr.c
```diff
@@ -0,0 +1,2610 @@
...
+			/* the chain of IXFRs ends in this serial number */
+			*end_serial = p->newserial;
+		}
+		p = next;
+	}
+	return 1;
+}
+
+/* Count length of next record in data */
+static size_t count_rr_length(uint8_t* data, size_t data_len, size_t current)
```
k0ekk0ek: Maybe `const` for `data`?

ixfr.c
```diff
@@ -0,0 +1,2612 @@
...
+		buffer_write_u8(packet, 0);
+		wirelen += 1;
+	}
+	return wirelen;
+}
+
+/* write an RR into the packet with compression for domain names,
+ * return 0 and resets position if it does not fit in the packet. */
+static int ixfr_write_rr_pkt(struct query* query, struct buffer* packet,
+	struct pktcompression* pcomp, uint8_t* rr, size_t rrlen)
```
k0ekk0ek: Probably best to make `rr` `const`?

TCY16 (approved):

k0ekk0ek (approved):
I've looked quite a bit at the PR. There's some places where the parameters/variables can be `const` (I'm sure I haven't marked all of them). But I haven't found anything that's obviously wrong and the tests pass. My mind is really not a compiler, so I can only trust all the buffer writes at certain offsets are correct. Personally, in places like that I'd combine some portions, e.g. maybe reuse the packet compression code that's currently already in and make it somewhat more generic. Then write a bunch of unit tests to test it more extensively. But I'm fine with it either way.

ixfr.c
wcawijngaards: Fixed in https://github.com/NLnetLabs/nsd/pull/209/commits/1454639924f6785a412c0e4577bbf1c8e6c6c14a .

ixfr.c
wcawijngaards: Fixed in https://github.com/NLnetLabs/nsd/pull/209/commits/b458e56e95641d074656b60f37519a4d5d8a00d0 . Thanks for the comments!

axfr.c
```diff
@@ -163,64 +168,78 @@ query_axfr(struct nsd *nsd, struct query *query)
...
+			addr2str(&q->addr, a, sizeof(a));
+			VERBOSITY(2, (LOG_INFO, "%s for %s from %s refused, %s",
+				(q->qtype==TYPE_AXFR?"axfr":"ixfr"),
+				dname_to_string(q->qname, NULL), a, acl?"blocked":"no acl matches"));
+		}
+		DEBUG(DEBUG_XFRD,1, (LOG_INFO, "%s refused, %s",
+			(q->qtype==TYPE_AXFR?"axfr":"ixfr"),
+			acl?"blocked":"no acl matches"));
+		if (!zone_opt) {
+			RCODE_SET(q->packet, RCODE_NOTAUTH);
```
wtoorop, line 191: Yes, I can see that we don't need an EDE code here because the rcode is unambiguous, but how is that different from here: https://github.com/NLnetLabs/nsd/blob/ixfr-out/query.c#L1299 . Should we not also set the NOAUTH rcode without the EDE there too? WDYT?
I know it is only moved code b.t.w. So I'm perfectly fine ignoring it. Just wanted to know your opinion.

ixfr.h
```diff
@@ -0,0 +1,270 @@
...
+	 * sum of sizes of the ixfr data elements */
+	size_t total_size;
+	/* the oldest serial number in the tree, searchable by old_serial */
+	uint32_t oldest_serial;
+	/* the newest serial number in the tree, that is searchable in the
+	 * tree, so it is the old_serial of the newest data entry, that
+	 * has an even newer new_serial of that entry */
+	uint32_t newest_serial;
+	/* the newserial number of the newest ixfr data element */
+	uint32_t newest_newserial;
```
wtoorop: What is it's purpose? It looks like values are assigned to it only, but never actually used. Is that correct?

ixfr.c
```diff
@@ -0,0 +1,2661 @@
...
+	size_t dname_len = dname_length(rr, rrlen);
+	if(!rr || rrlen == 0 || dname_len == 0)
+		return 0;
+	while(rrlen > 0 && rr[0] != 0) {
+		size_t lablen = (size_t)(rr[0]);
+		uint16_t offset;
+		if( (lablen&0xc0) )
+			break; /* name should be uncompressed */
+		if(lablen+1 > rrlen)
+			break; /* name should fit */
```
wtoorop: Should these error conditions not return 0 instead of breaking?

wtoorop (approved):
Thank you Wouter, looks wonderfull!

axfr.c
wcawijngaards: We set the REFUSED rcode in query.c because other implementations also return REFUSED for queries to a not hosted zone. And we want to be compatible.

ixfr.h
wcawijngaards: I added it because it allows comparison with the current serial in the nsd domain tree. But that comparison does not happen from that contents. I guess I should remove it, if it is not used.

ixfr.c
wcawijngaards: Fixed in https://github.com/NLnetLabs/nsd/pull/209/commits/166394f58efd6f6f41dd7d1f90906a43ac6c3877 . Good to find, even though wrong input is not expected here, the output is then nicer.

But after the commit, I think, what return 0 does, is return that it does not fit. And then it acts as if the RR did not fit. But instead we want to fail the transfer or something.

wcawijngaards: The fix in https://github.com/NLnetLabs/nsd/pull/209/commits/a954ac477c6ad5125033b52c83b62cf9ec615518 skips the RR if there is malformed dname content in it. It returns -1 to signal the malformed data.

ixfr.h
wcawijngaards: Fixed this in https://github.com/NLnetLabs/nsd/pull/209/commits/d675c33aa921bb319eae12526027f6d97fb2c0c5 . That removes the variable.

PR #210, opened 2022-03-24
Allow user to specify socket type per ip-addres

k0ekk0ek:
Listening on both UDP and TCP is likely unwanted behavior when support for XDP lands. Allow the user to configure listening on UDP, TCP or both per ip-address option in preparation. Specifying no socket type at all ensures both UDP and TCP are opened, which is the current default behavior, thereby remaining backwards compatible configuration wise.

### DIFF
--- a/nsd.c
+++ b/nsd.c
@@ -63,6 +63,9 @@ static char hostname[MAXHOSTNAMELEN];
 extern config_parser_state_type* cfg_parser;
 static void version(void) ATTR_NORETURN;
 
+static const char *default_tcp_port = TCP_PORT;
+static const char *default_udp_port = UDP_PORT;
+
 /*
  * Print the help text.
  *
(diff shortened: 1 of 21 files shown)


nsd.c
```diff
@@ -151,145 +154,400 @@ version(void)
...
+#else
+	assert(sockaddr->sa_family == AF_INET);
+#endif
+		addrlen = sizeof(struct sockaddr_in);
+	memmove(&socket->address.inet, sockaddr, addrlen);
+
+	setup_socket_servers(socket, ip);
+
+	if (ip && ip->dev) {
+		socket->flags = NSD_BIND_DEVICE;
```
wcawijngaards, line 409:
```suggestion
		socket->flags |= NSD_BIND_DEVICE;
```
Should this 'or' in the flag, because another flag exists, for OPTIONAL?

nsd.c
```diff
@@ -151,145 +154,400 @@ version(void)
...
 {
-	size_t i = 0, n = 1;
-	struct addrinfo ai[2] = { *hints, *hints };
+	int count = 0, exists = 0;
+	char name[IFNAMSIZ];
+	size_t namelen = 0;
+	unsigned short udp_port = 0, tcp_port = 0;
+
+	/* not an interface if length exceeds size restrictions */
+	if (!namelen || namelen >= IFNAMSIZ)
```
wcawijngaards, line 425: namelen is always 0, because initialized at 0 on line 421 above.

nsd.c
```diff
@@ -303,166 +561,72 @@ figure_default_sockets(
...
-			                   find_device(&(*tcp)[i], ifa) == 0))
-			{
-				error("cannot find device for ip-address %s",
-				      ip->address);
-			}
-		}
-#endif
+	for(ip = nsd.options->ip_addresses; ip; ip = ip->next) {
+		if (!setup_if_sockets(ip, ifaddrs))
+			setup_ip_sockets(ip, ifaddrs);
```
wcawijngaards, line 622: So the user can only specify all interface names, or all numeric ip addresses, but not mix them on individual interface config lines? Is that a correct reading of this part, not sure if users would need to do that.

CryptoGodfatherVA23 (approved):
Vedarius A. Russell

CryptoGodfatherVA23 (approved):

CryptoGodfatherVA23 (approved):

wcawijngaards (approved):
The changes look nice, it allows for more expressive configuration.

Issue #212, opened 2022-04-05
nsd-control verbosity 2 does not appear to take effect


anandb-ripencc:
I have a server started with verbosity 1 in the config:
```
server:
  verbosity: 1
```
While it is running, I want to see more details, so I use nsd-control to change verbosity. I expect the command to be logged, and the verbosity to change from 1 to 2. However, the command is not even logged, and I don't see any change in the amount of logging. Asking for the server status is logged, and shows the verbosity level as 2. When I later run `nsd-control verbosity 1`, this *does* get logged.
```
# nsd-control verbosity 2
ok
# nsd-control status
version: 4.4.0
verbosity: 2
ratelimit: 0
# nsd-control verbosity 1
# fgrep control /var/log/nsd/nsd.log
[2022-04-05 14:27:56.215] nsd[2181]: info: new control connection from /run/nsd/nsd.sock
[2022-04-05 14:27:56.215] nsd[2181]: info: control cmd:  status
[2022-04-05 14:33:50.747] nsd[2181]: info: new control connection from /run/nsd/nsd.sock
[2022-04-05 14:33:50.747] nsd[2181]: info: control cmd:  verbosity 1
```
This seems like a bug.

anandb-ripencc:
Okay, looking at the code, it looks like control commands are only logged at verbosity level 2. So the request to change verbosity from 1 to 2 won't be logged, but the request to lower it from 2 to 1 will get logged.

PR #215, opened 2022-05-04
Always log command line control actions

TCY16:
As discussed recently; change the verbosity of logging of command line controls to the lowest, so we always log command line control actions. Adresses #212.


wcawijngaards (approved):
Okay from me.

wtoorop (approved):
Prima

PR #218, opened 2022-06-30
Container support

jschlyter:
This PR implements initial support for building NSD as a container.

- The image is based on `debian:stable`.
- Directories for configuration and zones inspired by our `knot` friends.

This could be implemented in its own repo out of tree, but there might be benefits to keep it in sync with the releases.


jschlyter:
Published as https://github.com/jschlyter/nsd-container for now.

PR #219, opened 2022-07-03
Avoid c99 features to not reduce portability for no reason.

fobser:


k0ekk0ek:
Hi [@fobser](https://github.com/fobser). Thanks for the PR! Definitely willing to press the merge button, but before I do, I'd like to know the which compilers have issues with the original code. The reason I'm asking is:
1. I don't fully agree with "for no reason". Variable lifetime and having the compiler enforce it, is something I personally use to avoid mistakes/bugs. True, in these specific changes it's all pretty trivial, but anything I can use to avoid bugs, I usually do.
2. There's always the tradeoff between newer features and portability. c99 is 23 years old... If there's fairly modern compilers or operating systems that have problems with the code, then I'd like to take this as a learning opportunity. That way, in the future, I can at least make a conscious choice. Hopefully saving you the trouble of issuing a PR :slightly_smiling_face:

fobser:
I fully agree with with your point 1. So far I found it good enough to declare variables in the enclosing block rather than in the loop itself, i.e.
```
if (foo) {
    int i;

    for (i = 0; i < 42; i++) {
    }
}
```
I think this reduces the scope sufficiently. I guess what I'm saying is that so far I never ran into a non-trivial situation where the scope of the variable **must** be inside of the loop block.

About on which compilers this causes trouble, this does not compile on OpenBSD for the sparc64, alpha and HP PA-RISC architectures (gcc 4.2.1). As well as the LUNA-88K architecture (gcc 3.x). I think it would also not compile on RHEL6 which has gcc 4.4.x and is "only" EOL for 1.5 years. Things surely move slowly...

On OpenBSD nsd is part of the base system and must be compiled on all architectures with their native compiler.
All other OpenBSD supported architectures ( https://www.openbsd.org/plat.html ) use llvm/clang.

k0ekk0ek:
Thanks for your comment @fobser! Judging by this [table](https://gcc.gnu.org/c99status.html), the constructs should be supported by every compiler you mention. I want to see if I can get it to work for a bit. Should I fail, I'll happily merge the PR.

fobser:
Never mind, I have just been told that `-std=c99` should fix the issue. The problem is, those compilers don't do c99 by *default*. Sorry for the noise.
I'll close this, my mistake.

k0ekk0ek:
That's just what I was about to try out to update our makefile :slightly_smiling_face: (Trying to compile GCC 3.0.x on a modern system isn't the easiest :wink:) Glad it works.

k0ekk0ek:
Using `AC_PROG_CC_STDC` now. @fobser, can I ask you confirm that makes adding the flag automatic for you too?

fobser:
This is on `sparc64` with `gcc version 4.2.1`
```
checking for cc option to accept ISO C99... -std=gnu99
[...]
config.status: creating config.h
cc -std=gnu99 -I. -I/usr/src/usr.sbin/nsd -O2 -pipe   -c /usr/src/usr.sbin/nsd/answer.c
[...]
```

I'll ask the `LUNA-88K` maintainer to test on gcc3, too. Might take a while.
From what I hear from our compiler experts, it should just work.

fobser:
It also works with gcc3

k0ekk0ek:
Thanks!

PR #225, opened 2022-08-16
Add SVCB dohpath support

TCY16:


rdata.c
```diff
@@ -798,6 +798,22 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+rdata_svcparam_dohpath_to_string(buffer_type *output, uint16_t val_len,
+	uint16_t *data)
+{
+	uint8_t *dp = (void *)data;
+
+	assert(val_len > 0); /* Guaranteed by rdata_svcparam_to_string */
+
+	buffer_write_u8(output, '=');
+	buffer_write_u8(output, '"');
+	buffer_printf(output, "%s", dp);
```
wcawijngaards: Please some type of isprint check on characters, so that binary data results in printable text on the output. Like, by escaping the characters if they are not printable. Like in rdata_text_to_string(), the characters are escaped if needed.

zonec.c
```diff
@@ -1132,6 +1136,7 @@ zparser_conv_svcbparam_key_value(region_type *region,
 		return zparser_conv_svcbparam_ech_value(region, val);
 	case SVCB_KEY_ALPN:
 		return zparser_conv_svcbparam_alpn_value(region, val, val_len);
+	case SVCB_KEY_DOHPATH:
```
wcawijngaards, line 1139: If there is output of escapes for non printable characters, here is where it then needs to be unescaped, I guess.

wcawijngaards (approved):
The code looks good. The output of wire to string, I would like it if binary content was turned into printable text output. With like an isprint check.

zonec.c
TCY16: This isn't used in the other "simple" SVCB params either

rdata.c
```diff
@@ -798,6 +798,29 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+			buffer_printf(output, "\\%03u", (unsigned) dp[i]);
+		} else {
+			buffer_printf(output, "%c", dp[i]);
+		}
+	}
+	buffer_write_u8(output, '"');
+
+	return 1;
+}
+
```
wtoorop:
```suggestion
```

Together with suggestion below this would be more consistent (because now quotes and backslashes are not escaped)

rdata.c
```diff
@@ -844,6 +868,8 @@ rdata_svcparam_to_string(buffer_type *output, rdata_atom_type rdata,
 		return rdata_svcparam_alpn_to_string(output, val_len, data+2);
 	case SVCB_KEY_ECH:
 		return rdata_svcparam_ech_to_string(output, val_len, data+2);
+	case SVCB_KEY_DOHPATH:
+		return rdata_svcparam_dohpath_to_string(output, val_len, data+2);
```
wtoorop:
```suggestion
	case SVCB_KEY_DOHPATH:
		/* fallthrough */
```

zonec.c
wtoorop, line 1139:
```suggestion
	case SVCB_KEY_DOHPATH:
		/* fallthrough */
```

wtoorop (changes requested):
Remove some redundant code... and make it more consistent quote-wise

zonec.c
wtoorop: An unescaped value is already passed to `zparser_conv_svcbparam_key_value()` so it doesn't need to be unsecaped here again (except with alpn values which are double-escaped).

Issue #227, opened 2022-08-17
IXFR serial quirks when NSD restarts and only have old zone file on disk


klaus-nicat:
Setup: NSD 4.6 (compiled with --disable-radix-tree and --enable-packed).

```
server:
        # the database to use
        # database: "/dns/nsd/nsd.db"
        database: ""

        # Number of NSD servers to fork.
        server-count: 1

        # Verbosity level.
        verbosity: 2

        # The directory for zonefile: files.
        zonesdir: "/var/lib/nsd/"

zone:
        name: example.net
        zonefile: %s
        allow-axfr-fallback: yes
        allow-notify: 136.244.78.244 NOKEY
        request-xfr:  136.244.78.244 NOKEY
        max-refresh-time: 300
```

NSD is Secondary. Serial on Primary is 1, Serial on NSD is 1.

```
root@klaus-nsd-slave:~# nsd-control zonestatus
zone:   example.net
        state: ok
        served-serial: "1 since 2022-08-17T20:00:26"
        commit-serial: "1 since 2022-08-17T20:00:26"
        wait: "292 sec between attempts"
```

Then, write the zonefile:
```
# nsd-control write
```

Now, upate the zone on the Primary to serial=2. NSD will IXFR the new zone, but does not write it to disk:
```
Aug 17 20:05:43 klaus-nsd-slave nsd[15797]: xfrd: zone example.net committed "received update to serial 2 at 2022-08-17T20:05:43 from 136.244.78.244"
Aug 17 20:05:43 klaus-nsd-slave nsd[15810]: zone example.net. received update to serial 2 at 2022-08-17T20:05:43 from 136.244.78.244 of 235 bytes in 3.8e-05 seconds
Aug 17 20:05:43 klaus-nsd-slave nsd[15797]: zone example.net serial 1 is updated to 2

# nsd-control zonestatus
zone:   example.net
        state: ok
        served-serial: "2 since 2022-08-17T20:05:43"
        commit-serial: "2 since 2022-08-17T20:05:43"
        wait: "281 sec between attempts"
```

Now restart NSD. NSD will serve the old version with serial=1, as this is the latest version on disk.
```
Aug 17 20:07:33 klaus-nsd-slave nsd[15861]: nsd started (NSD 4.6.0), pid 15845
Aug 17 20:07:33 klaus-nsd-slave nsd[15845]: zone example.net serial 2 is updated to 1

# nsd-control zonestatus
zone:   example.net
        state: ok
[...]

PR #229, opened 2022-08-18
Not the fix for Issue #227

wtoorop:
Not the fix because problem became worse (see Issue #229), but it is a first step.
TODO: create a tpkg test and work from there making a proper fix.


wtoorop:
Closing this in favor of PR #238

PR #230, opened 2022-08-18
Change zone parsing to accept non-trailing newline

TCY16:


k0ekk0ek:
So, basically what this change does is not make the zone parser require a newline to delimit the record. I'm ok with it. I don't particulary like the static variable, but the parser isn't thread-safe anyway, so I'd be fine with this getting merged. @wcawijngaards, @wtoorop, any comments?

wcawijngaards:
I do not see any issues with it, if that works for non-newline-ended files, that is nice.

k0ekk0ek:
@TCY16, can you update the changelog too? After that, I think we can go ahead and merge this.

k0ekk0ek:
@TCY16, Changelog?

Issue #199, opened 2021-11-08
"dot" ALPN token MUST be negotiated for XoT (according to RFC9103, Section 7.1)


arbv:
The XoT specification [requires](https://www.rfc-editor.org/rfc/rfc9103.html#name-connection-establishment) `dot` ALPN token to be negotiated for zone transfers:

> 7.1. Connection Establishment
> During connection establishment, the Application-Layer Protocol Negotiation (ALPN) token "dot" [DoT-ALPN] MUST be selected in the TLS handshake.

At the very least, version 4.3.8 of NSD does not appear to do that, which on practice leads to interoperability problems with the current development release of BIND (9.17.19) and `dig`, which require that for zone transfers according to the spec ([details](https://gitlab.isc.org/isc-projects/bind9/-/merge_requests/5391)).

cesarkuroiwa:
We currently have Bind as a primary to some NSDs as secondaries, and wish to deploy XoT, but this issue is holding us back

k0ekk0ek:
Thanks for the report @arbv. Sorry this didn't get picked up sooner, but luckily @cesarkuroiwa helped out :sweat_smile: I'll merge his PR ASAP after which I'll close this issue. Thanks to both of you :+1:

arbv:
I am glad to hear that the problem is solved! :+1:

arbv:
There is a catch, though. Judging from the MR, the case when NSD acts as a primary and BIND as the secondary is still not supported as no ALPN happens on accepted connections.

k0ekk0ek:
Let's reopen this one then. I'll have a look at the server side, see how easy/hard it'd be to get that in.

wtoorop:
This has been done in commit e6a1024 which has been released as NSD 4.11.1 released on January 18

PR #234, opened 2022-09-27
Set 'dot' as an available ALPN protocol

cesarkuroiwa:
RFC9103, Section 7.1 

#199


k0ekk0ek:
Hi @cesarkuroiwa. Thanks! Looks like a good addition. I've taken the liberty of moving the protocol list to the start of the function and update the changelog. If the checks pass, I'll hit merge :+1: Thanks again!

PR #238, opened 2022-11-03
Fix serial quirks for IXFR after restart

k0ekk0ek:
For details, see #227. This PR adds a test to verify the correct serial is used for IXFRs after a restart and a fix for the issue reported.

### DIFF
--- a/tpkg/ixfr_and_restart.tdir/ixfr_and_restart.conf
+++ b/tpkg/ixfr_and_restart.tdir/ixfr_and_restart.conf
@@ -1,0 +1,25 @@
+server:
+  logfile: "nsd.log"
+  xfrdir: ""
+  xfrdfile: "xfrd.state"
+  zonelistfile: "zone.list"
+  database: ""
+  pidfile: "nsd.pid"
+  server-count: 1
+  port: NSD_PORT
+  verbosity: 5
+  zonesdir: ""
+  username: ""
+  chroot: ""
+
+remote-control:
+  control-enable: yes
+  control-interface: NSD_CTRL_PATH
+
+zone:
+  name: example.net
+  zonefile: %s
+  allow-axfr-fallback: yes
+  allow-notify: 127.0.0.1 NOKEY
+  request-xfr: 127.0.0.1@TESTNS_PORT NOKEY
+  max-refresh-time: 300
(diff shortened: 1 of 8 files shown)


tpkg/ixfr_and_restart.tdir/ixfr_and_restart.dsc
```diff
@@ -0,0 +1,16 @@
+BaseName: ixfr_and_restart
+Version: 1.0
+Description: Verify NSD requests IXFR with correct serial after restart
+CreationDate: Thu Nov 01 14:11:27 CET 2020
```
wcawijngaards:
```suggestion
CreationDate: Thu Nov 01 14:11:27 CET 2022
```
Typing error fixed. The date is not used to run the test framework, anyway.

tpkg/ixfr_and_restart.tdir/ixfr_and_restart.conf
```diff
@@ -0,0 +1,24 @@
+server:
+  logfile: "nsd.log"
+  xfrdir: ""
+  xfrdfile: xfrd.state
```
wcawijngaards:
```suggestion
  xfrdfile: xfrd.state
  zonelistfile: "zone.list"
```
The test should specify a zonelistfile. The test does not use it, but it is neater to not refer to the system default zone list file from the unit test.

wcawijngaards (approved):
The changes look good.
It would be good to also fix xfrd.c:2359 compare_serial call parameters.

tpkg/ixfr_and_restart.tdir/ixfr_and_restart.dsc
k0ekk0ek: Included

tpkg/ixfr_and_restart.tdir/ixfr_and_restart.conf
k0ekk0ek: Included

k0ekk0ek:
> The changes look good. It would be good to also fix xfrd.c:2359 compare_serial call parameters.

Included this one too, good catch :+1:

klaus-nicat:
Is this ready for testing?

k0ekk0ek:
Hi @klaus-nicat, yes, it is.

wtoorop (approved):
Excellent! Thanks!

PR #241, opened 2022-12-31
get rid of the PRIu32 define

janstary:
This fixes the build on macOS Ventura 13.1;
the three occurences of PRIu32 can be replaced
by the obvious %u without losing anything.

### DIFF
--- a/xfrd.c
+++ b/xfrd.c
@@ -1322,10 +1322,10 @@ xfrd_handle_incoming_soa(xfrd_zone_type* zone,
 			soa_serial = ntohl(soa->serial);
 			soa_nsd_serial = ntohl(zone->soa_nsd.serial);
 			if (compare_serial(soa_serial, soa_nsd_serial) > 0)
-				log_msg(LOG_INFO, "zone %s serial %"PRIu32" is updated to %"PRIu32,
+				log_msg(LOG_INFO, "zone %s serial %u is updated to %u",
 					zone->apex_str, soa_nsd_serial, soa_serial);
 			else
-				log_msg(LOG_INFO, "zone %s serial is updated to %"PRIu32,
+				log_msg(LOG_INFO, "zone %s serial is updated to %u",
 					zone->apex_str, soa_serial);
 		}
 		zone->soa_nsd = *soa;


k0ekk0ek:
Hi @janstary, Thanks for testing and the associated patch!

I'd prefer to keep using `PRIu32` as we're really printing an `uint32_t` rather than `unsigned int`. While they're usually the same size, that doesn't necessarily need to be the case. Usually, including `<inttypes.h>` should solve the problem, can you give that a try?

If that does the trick, eventually, we may have to check if that header is available in autoconf and optionally include it in the generated `config.h` so that we don't break builds on older systems(?) Feel free to tackle that part too, but as that's slightly more work, I'm happy to take care of it too.

janstary:
`PRIu32` is defined as `u` in `inttypes.h`; can you please give an example where it is not?
Including a file just to be able to say `PRIu32` instead of `u` is the opposite of my intention to simplify.

```
$ grep -Fr %u * | wc -l
     158
```

There's 158 occurences of `printf("%u")` and 3 occurences of `printf("%"PRIu32)` in the code;
see for example line 662 of the same file where the serial is printed as `%u` (obviously).

What I propose is to get rid of this, not to accommodate it.

janstary:
Having configure check whether inttypes.h exists is another needless complication.
The whole thing is pointless AFAICT: on what system is printf("%u", serial) a problem
in case `uint32_t` is not an `unsigned int`? (And where is that the case?)

k0ekk0ek:
All common platforms are very likely to just typedef `uint32_t` to `unsigned int` and therefore `PRIu32` is also just defined to `u` in the formatting flags. However, `uint32_t` isn't necessarily a typedef of `unsigned int`. While I have only encountered it for embedded boards (Xilinx Zync), where `uint32_t` and `unsigned int` where of the same size, but not the same type, it may be possible that other platforms/toolchains have that too. The use of `PRIu32` is actually part of a broader push (from me) to require at least C99 and fix problems as they turn up. We/I simply haven't gotten around to replacing all instances, in large part because we/I don't want to change everything all at once. Therefore, all other instances should actually, technically, also be replaced by `PRIu32`. At the very least, I don't want to revert any changes that get us closer to requiring a newer specification when the problem at hand is easily solvable in another way, as is the case here. Not because I care so much about the integer flags in this particular case, but rather so we learn what features we can safely use that the newer specification introduced.

While I get that you just want it to get it to compile, for the project at large, I think it's better that we spend a little bit of extra time.

As for the header check, you're probably right and most platforms just provide it. However, builds on platforms that practically provide C99, but not "C99" may break if we just include it. There's at least one build that still depends on GCC 3.3 (I believe), that provides "gnu99" but not "C99". We'd have to check if `inttypes.h` can safely be included and make it optional (or provide our own alternative) if it breaks builds.

Other maintainers may not feel as strongly about this btw. Looking at the bigger picture though, I hope you can see why I prefer to keep `PRIu32`.

janstary:
I still don't understand what problem the PRIu32 is trying to solve.
What exactly happens with e.g. `printf("%u", serial)` when `serial`, which is an `uint32_t`,
happens to be something else than an `unsigned int`? What else would it be, and what problem
would it mean for the printf? What would PRIu32 be on such a platform,
and how exactly would that solve that probem?

janstary:
What build depends on GCC 3.3 and in what way does it _practically provide C99, but not "C99"_?

k0ekk0ek:
It's relatively simple to explain: The C specification does not give guarantees about the size of integer types. On all modern CPU's, `int`s are very likely 32-bits. But they don't have to be. The `printf` flag `%u` is used to print `unsigned int`. C99 added fixed integer types, e.g. `uint32_t`. The proper flag to print an `uin32_t` is `PRIu32`.

The platform requiring GCC3.3 is the LUNA-88K port of OpenBSD, see #219.

What I mean by _practically provide C99, but not C99_, is that a compiler may support only a subset. https://gcc.gnu.org/c99status.html has a nice detailed overview. You'll notice that 3.3 supports a lot, but not everything, or with slightly different semantics.

What exactly happens in `printf` would be guessing. If they're the same size (not type) and the compiler applies the same rules to both, everything could work out, but you'll get lots of compiler warnings.

janstary:
> It's relatively simple to explain: The C specification does not give guarantees about the size of integer types. On all modern CPU's, `int`s are very likely 32-bits. But they don't have to be. The `printf` flag `%u` is used to print `unsigned int`. C99 added fixed integer types, e.g. `uint32_t`. The proper flag to print an `uin32_t` is `PRIu32`.

Sorry, but that does not explain anything: what is the actual problem? If `unsigned int` happens to be 16 bit,
while the `serial` variable is `uint32_t`,  what is the problem with `printf("%u", serial)`?
What will PRIu32 be on such a platform, and how will `printf("%"PRIu32, serial)` solve the problem?

janstary:
On my (only) 16bit i386 machine,

```
#define PRIu16                  "u"             /* uint16_t */
#define PRIu32                  "u"             /* uint32_t */
```

So yeah, it doesn't do anything really.

k0ekk0ek:
@janstary, there's one instance that I've encountered (mentioned above) where the type was of the same size but not a typedef of unsigned int which led to a lot of warnings. Sorry I'm unable to explain this in more detail.

Just wanted to close this one and let you know I've issued a PR that should fix the issue. It appears OpenBSD 7.2 suffered from the same problem and I've updated the build environment to hopefully spot things like this next time. Thanks for issuing the report, sorry I wasn't able to accept your PR.

janstary:
So what is `uint32_t` on the Xilinx Zync, where `uint32_t` and `unsigned int` are of the same size,
but not the same type? What is PRIu32 defined to be on this platform? And what is the problem that
happens with `printf("%u", serial)` as `serial` is a `uint32_t` but not an `unsigned int`?

k0ekk0ek:
I actually tried to find the cross compile toolchain when composing my previous reply. I suppose I can still do that, but downloading and installing the SDK takes ages. They used to have just the compiler available, but after spending an hour trying to find it, I gave up. What I could dig up was (one of) the [commits](https://github.com/eclipse-cyclonedds/cyclonedds/commit/c9d827e420b7074a9b319f2faaddae09ed0d961f) that fixed a bunch of compiler warnings for that particular build. Here's [one more](https://github.com/eclipse-cyclonedds/cyclonedds/commit/a65312d34ec3022e95cb9960a133c1c5d589897a) from a colleague at the time that outlines it was specifically for a FreeRTOS build (that information wasn't clear from the first).

`uint32_t` is usually an alias for `unsigned int`, in which case the `%u` flag can safely be used because it's both apples. If, however, `uint32_t` is not an alias for `unsigned int`, that no longer holds, because it may be apples and oranges.

That in itself should be enough to prefer the fix proposed in #255.

As previously stated, I don't particularly care about the `printf` flags. I'll admit that for general purpose hardware, they're very likely to always be the same. However, it's not the flags that make me prefer the alternate fix, it's the fact that I (eventually) want to just require proper c99 support to build NSD. In order to do that, I need to make sure all platforms in use support that specification and fix problems where they pop up. We managed to make it work with GCC 3.3 for OpenBSD on the [LUNA-88K2](https://www.openbsd.org/luna88k.html), so it seems silly to regress for brand new platforms where we're just missing an include.

That's really as clear as I can state it without burning precious amounts of time trying to find the (or a, for that matter) compile toolchain and show the actual defines.

PR #255, opened 2023-01-06
Include inttypes.h

k0ekk0ek:
This PR includes `inttypes.h` in `xfrd.c` to fix the build issue reported in #241. I wasn't able to reproduce it on macOS, but oddly enough it occurred on OpenBSD 7.2 too. I've purposely left out checking for `inttypes.h` in the configure script (for now), to see if any build errors are reported before the next release in trying to gain knowledge about what we can reasonably require in terms of C99 compatibility. It goes without saying that should any build issues be reported because of `inttypes.h` missing on that platform the configure check will be retrofitted.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -7,18 +7,15 @@
 # For now builds are limited to Linux, FreeBSD and macOS builds on AMD64, more
 # may be added in the future.
 
-ubuntu_2004_gcc11: &ubuntu_2004_gcc11
+ubuntu_2204_gcc11: &ubuntu_2204_gcc11
   container:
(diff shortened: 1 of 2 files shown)


k0ekk0ek:
Turns out, Cirrus CI has macOS Ventura available and the macOS 10.15 images were deprecated from January 1st 2023, so I went ahead and just updated all systems used.

wcawijngaards (approved):
The ci improvement for newer platform versions looks good. The include code line looks fine.

Maybe that works for everyone, people have fairly new systems, that use NSD.

PR #260, opened 2023-01-23
Fix two ignored return values

wtoorop:
That were reported when compiling:
```
server.c: In function ‘server_verify’:
server.c:3037:9: warning: ignoring return value of ‘pipe’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
 3037 |         pipe(nsd->verifier_pipe);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~

verify.c: In function ‘verify_handle_signal’:
verify.c:277:15: warning: ignoring return value of ‘write’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
  277 |         (void)write(nsd->verifier_pipe[1], buf, sizeof(buf));
      |               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/server.c
+++ b/server.c
@@ -3034,7 +3034,11 @@ void server_verify(struct nsd *nsd, int cmdsocket)
 	nsd->verifier_count = 0;
 	nsd->verifier_limit = nsd->options->verifier_count;
 	size = sizeof(struct verifier) * nsd->verifier_limit;
-	pipe(nsd->verifier_pipe);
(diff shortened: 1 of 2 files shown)


k0ekk0ek:
I think for the change in `server_verify`, we should probably add another label that skips closing the pipe(?) Maybe add a `fail_pipe` label after `fail` that skips closing the descriptors? Other than that I'm fine with the changes.

PR #263, opened 2023-01-29
Add bash autocompletion script for nsd-control

cweiske:
It completes options, all commands and zone arguments for commands.

To use it, call:

    source contrib/autocomplete_nsd-control.bash


wcawijngaards (approved):
This is a nice addition for the contrib section, I tried it briefly and it works.

wcawijngaards:
Thank you for the additional commandline support! The file has been added to the code repository.

PR #265, opened 2023-02-01
Fix C99 compatibility issue

djdelorie:
Two of the configure tests need prototypes for close() to avoid C99 warnings.

Related to:

      <https://fedoraproject.org/wiki/Changes/PortingToModernC>
      <https://fedoraproject.org/wiki/Toolchain/PortingToModernC>


wcawijngaards (approved):
This change looks good, there do not seem to be other implicit function declaration warnings in the config output.

wcawijngaards:
Thank you for the patch! It works and removes the unneeded warnings from the configure logs. The changes have been merged with the code repository.

PR #266, opened 2023-02-09
Fix build with --without-ssl

cotequeiroz:
Don't include COMMON_OBJ when compiling nsd-control without SSL.

Without SSL, nsd-control's only purpose is to print out an error message.  There's no need to include any other object files then.

Signed-off-by: Eneas U de Queiroz <cotequeiroz@gmail.com>

----

Here's how the build fails with OpenWrt:

```
aarch64-openwrt-linux-musl-gcc -O3 -pipe -mcpu=cortex-a53 -funsafe-math-optimizations -fno-plt -fhonour-copts -fmacro-prefix-map=/home/equeiroz/src/openwrt/build_dir/target-aarch64_cortex-a53_musl/nsd-nossl/nsd-4.6.1=nsd-4.6.1 -Wformat -Werror=format-security -fstack-protector -D_FORTIFY_SOURCE=1 -Wl,-z,now -Wl,-z,relro  -Wno-address-of-packed-member -L/home/equeiroz/src/openwrt/staging_dir/toolchain-aarch64_cortex-a53_gcc-12.2.0_musl/usr/lib -L/home/equeiroz/src/openwrt/staging_dir/toolchain-aarch64_cortex-a53_gcc-12.2.0_musl/lib -znow -zrelro  -o nsd-control answer.o axfr.o ixfr.o ixfrcreate.o buffer.o configlexer.o configparser.o dname.o dns.o edns.o iterated_hash.o lookup3.o namedb.o nsec3.o options.o packet.o query.o rbtree.o radtree.o rdata.o region-allocator.o rrl.o siphash.o tsig.o tsig-openssl.o udb.o udbradtree.o udbzone.o util.o bitset.o popen3.o nsd-control.o memcmp.o cpuset.o b64_pton.o b64_ntop.o setproctitle.o
/home/equeiroz/src/openwrt/staging_dir/toolchain-aarch64_cortex-a53_gcc-12.2.0_musl/lib/gcc/aarch64-openwrt-linux-musl/12.2.0/../../../../aarch64-openwrt-linux-musl/bin/ld: ixfr.o: in function `ixfr_data_readrr':
ixfr.c:(.text+0x370): undefined reference to `zonec_parse_string'
collect2: error: ld returned 1 exit status
make[3]: *** [Makefile:167: nsd-control] Error 1
```

This was compile-tested with OpenWrt both without SSL and with OpenSSL 3.0.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -85,7 +85,7 @@ NSD_OBJ=$(COMMON_OBJ) $(XFRD_OBJ) difffile.o ipc.o mini_event.o netio.o nsd.o se
 ALL_OBJ=$(NSD_OBJ) nsd-checkconf.o nsd-checkzone.o nsd-control.o nsd-mem.o xfr-inspect.o
 NSD_CHECKCONF_OBJ=$(COMMON_OBJ) nsd-checkconf.o
 NSD_CHECKZONE_OBJ=$(COMMON_OBJ) $(XFRD_OBJ) dbaccess.o dbcreate.o difffile.o ipc.o mini_event.o netio.o server.o zonec.o zparser.o zlexer.o nsd-checkzone.o verify.o
-NSD_CONTROL_OBJ=$(COMMON_OBJ) nsd-control.o
+NSD_CONTROL_OBJ=@NSD_CONTROL_COMMON_OBJ@ nsd-control.o
 CUTEST_OBJ=$(COMMON_OBJ) $(XFRD_OBJ) dbaccess.o dbcreate.o difffile.o ipc.o mini_event.o netio.o server.o verify.o zonec.o zparser.o zlexer.o cutest_dname.o cutest_dns.o cutest_iterated_hash.o cutest_run.o cutest_radtree.o cutest_rbtree.o cutest_namedb.o cutest_options.o cutest_region.o cutest_rrl.o cutest_udb.o cutest_udbrad.o cutest_util.o cutest_bitset.o cutest_popen3.o cutest_iter.o cutest_event.o cutest.o qtest.o
 NSD_MEM_OBJ=$(COMMON_OBJ) $(XFRD_OBJ) dbaccess.o dbcreate.o difffile.o ipc.o mini_event.o netio.o verify.o server.o zonec.o zparser.o zlexer.o nsd-mem.o
 all:	$(TARGETS) $(MANUALS)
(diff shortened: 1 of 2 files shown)


cotequeiroz:
It appears this breaks the FreeBSD build.  Does it not set `HAVE_SSL` in the Makefile?  An alternative--I have it ready--is to move `zonec_parse_string()` and its header outside of the `HAVE_SSL` guard.

cotequeiroz:
I finally got it.  FreeBSD is using BSD make, not GNU make.  

I push a commit making `configure` responsible for adding `$(COMMON_OBJ)` to `nsd-control` prerequisites in the Makefile.

k0ekk0ek:
Hi @cotequeiroz! Thanks for giving this a try! Also, sorry for the overdue reply... `nsd-control` has function without OpenSSL as it allows connecting to a unix-domain socket for local operation. I'll have to look into this a bit further, but it seems acceptable to not require SSL in that case. I'm not sure if that's currently optional within NSD itself.

If my assumptions are correct: compiling without `COMMON_OBJ` leaves out at least `configlexer.o` and `configparser.o`, which are required to read the configuration file. I'm actually surprised `nsd-control` builds at all without OpenSSL, because the `setup_ssl` function is not within a conditional preprocessor block :sweat_smile:

We definitely have to look into building without SSL because there's some oddities. I'll have a good look and see if I can do a non-ssl build and go from there.

k0ekk0ek:
Hi @cotequeiroz, @wcawijngaards fixed the build issue in the commit referenced above. That should help you along. Not as nice as allowing local operation without SSL yet, but we can tackle that at a later stage. Thanks for taking the time for reporting and trying to come up with a fix. Much appreciated!

cotequeiroz:
Thanks.  I suggest you add a CI build without ssl. 
 https://github.com/cotequeiroz/nsd/commit/e3cdf658ba15233a8887d9337cadce8e7918d174

cotequeiroz:
> If my assumptions are correct: compiling without `COMMON_OBJ` leaves out at least `configlexer.o` and `configparser.o`, which are required to read the configuration file. I'm actually surprised `nsd-control` builds at all without OpenSSL, because the `setup_ssl` function is not within a conditional preprocessor block 😅
> 

All that was left outside of the `HAVE_SSL` guard was `main()`, which just printed out an error message and exited.  That's why I had excluded `COMMON_OBJ` completely.
https://github.com/NLnetLabs/nsd/blob/64595dabf529fe7d29a77ff0023ef0c99b91ffa0/nsd-control.c#L607-L613

PR #269, opened 2023-02-19
Add systemd service unit

Fale:
There was a fairly complex systemd unit that got deleted by its author in commit e502f434d0ca54c1ac83fba84339e822540e63f0. The commit did not removed the `nsd.socket` that got created together and did not updated the README file, so I did so.

Differently from the old systemd unit, this one is way simpler. It might be improved over time, but I think it is a good starting point.

### DIFF
--- a/contrib/README
+++ b/contrib/README
@@ -28,8 +28,7 @@ USE AT YOUR OWN RISK.
   You may also need to create a number of symbolic links under the names
   of the graphs you want to create (documented at head of file).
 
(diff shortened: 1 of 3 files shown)


wcawijngaards (approved):
This looks like a very good systemd file for NSD! It could be helpful for setting up the server integration with systemd after installing.

wcawijngaards:
Thank you for the contributed service script! It has been added to the code repository.

PR #276, opened 2023-04-03
Drop root-server configuration option

### LINKED ISSUE #275: enable-root-server as config option
Has there been any discussion on moving the compile time option `--enable-root-server` to a runtime configuration option? Where I understand why measures should be taken to protect users from using NSD as a root server, requiring a custom build seems a bit to complicated. Having this as a runtime configuration option would help a lot running of the shelf binary packages.

k0ekk0ek:
Fixes #275.

### DIFF
--- a/verify.c
+++ b/verify.c
@@ -112,11 +112,11 @@ static inline size_t print_line(struct verifier_stream *stream, int eof)
 		return 0;
 
 	if (len > LOGLINELEN) {
-		fmt = stream->cut ? ".. %.*s .." : "%.*s ..";
+		fmt = stream->cut ? "verifier: .. %.*s .." : "verifier: %.*s ..";
 		len = LOGLINELEN; // remainder printed next iteration
 		stream->cut = 1;
 	} else {
-		fmt = stream->cut ? ".. %.*s" : "%.*s";
+		fmt = stream->cut ? "verifier: .. %.*s" : "verifier: %.*s";
 		stream->cut = 0;
 	}
 	log_msg(stream->priority, fmt, len, stream->buf + stream->off);
--- a/configure.ac
+++ b/configure.ac
@@ -938,14 +938,7 @@ AC_DEFINE_UNQUOTED([TCP_TIMEOUT], $tcp_timeout, [Define to the default tcp timeo
 dnl
 dnl Features
 dnl
-AC_ARG_ENABLE(root-server, AS_HELP_STRING([--enable-root-server],[Configure NSD as a root server]))
-case "$enable_root_server" in
-        yes)
-		AC_DEFINE_UNQUOTED([ROOT_SERVER], [], [Define this to configure as a root server.])
-		;;
-        no|*)
-                ;;
-esac
+AC_ARG_ENABLE(root-server, AS_HELP_STRING([--enable-root-server],[Configure NSD as a root server (obsolete)]))
 
 AC_ARG_ENABLE(ipv6, AS_HELP_STRING([--disable-ipv6],[Disables IPv6 support]))
 case "$enable_ipv6" in
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,4 +1,5 @@
 30 May 2023: Jeroen
+	- Fix #240: Prefix messages originating from verifier.
 	- Fix #275: Drop unnecessary root server checks.
 
 30 May 2023: Wouter
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,3 +1,6 @@
+30 May 2023: Jeroen
+	- Fix #275: Drop unnecessary root server checks.
+
 30 May 2023: Wouter
 	- Next version is 4.7.0, instead of 4.6.2, because of the added
 	  features, like TLS for DNSTAP.
--- a/nsd-checkconf.c
+++ b/nsd-checkconf.c
@@ -791,13 +791,6 @@ additional_checks(nsd_options_type* opt, const char* filename)
 			errors ++;
 			continue;
 		}
-#ifndef ROOT_SERVER
-		/* Is it a root zone? Are we a root server then? Idiot proof. */
-		if(dname->label_count == 1) {
-			fprintf(stderr, "%s: not configured as a root server.\n", filename);
-			errors ++;
-		}
-#endif
 		if(zone->pattern->allow_notify && !zone->pattern->request_xfr) {
 			fprintf(stderr, "%s: zone %s has allow-notify but no request-xfr"
 				" items. Where can it get a zone transfer when a notify "
--- a/zonec.c
+++ b/zonec.c
@@ -2209,14 +2209,6 @@ zonec_read(const char* name, const char* zonefile, zone_type* zone)
 		return 1;
 	}
 
-#ifndef ROOT_SERVER
-	/* Is it a root zone? Are we a root server then? Idiot proof. */
-	if (dname->label_count == 1) {
-		zc_error("not configured as a root server");
-		return 1;
-	}
-#endif
-
 	/* Open the zone file */
 	if (!zone_open(zonefile, 3600, CLASS_IN, dname)) {
 		zc_error("cannot open '%s': %s", zonefile, strerror(errno));


dbaccess.c
```diff
@@ -499,9 +499,14 @@ namedb_read_zonefile(struct nsd* nsd, struct zone* zone, udb_base* taskudb,
...
 	struct ixfr_create* ixfrcr = NULL;
-	int ixfr_create_already_done = 0;
+	int ixfr_create_already_done = 0, root_server;
 	if(!nsd->db || !zone || !zone->opts || !zone->opts->pattern->zonefile)
 		return;
+#ifdef ROOT_SERVER
+	root_server = nsd->options ? nsd->options->root_server : 1;
+#else
+	root_server = nsd->options ? nsd->options->root_server : 0;
+#endif
```
wcawijngaards: Having the define to allow root-server by default makes sense.

The define here seems to be only for when nsd.options is NULL. This only happens during the unit test. So, it does not set the default here, but only the value used during unit tests. Perhaps it is cleaner to say so in a comment, or no have the if on nsd.options NULL or not, and instead make changes in the unit tests to set the value. tpkg/cutest/cutest_namedb.c:90 in create_and_read_db. nsd.options = opt;
tpkg/cutest/cutest_iter.c:56 in create_zone, nsd.options = options;  
And then also set nsd.options->root_server = 1; if it needs to be enabled for the unit test.

So, in that case, the code turns into ifdef ROOT_SERVER root_server = 1; else root_server = nsd->options->root_server; endif. Or am I missing the intent of the code and ifdef and NULL checks?

nsd-checkzone.c
```diff
@@ -80,7 +80,7 @@ check_zone(struct nsd* nsd, const char* name, const char* fname, FILE *out,
 	}
 
 	/* read the zone */
-	errors = zonec_read(name, fname, zone);
+	errors = zonec_read(name, fname, zone, nsd->options->root_server);
```
wcawijngaards: The value of nsd->options->root_server here is the default. The nsd.conf file has not been read in. For the user with a root zone that tries to nsd-checkzone the root.zone file, that would fail if they used the nsd.conf root-server: yes option. Because the default of no is from the options allocation routine, and the config file is not read at this point. It would need to read the config file to have that option, but this may also not be desirable, because of all the other stuff that can be in a config file, or maybe add a commandline option to set root-server value to nsd-checkzone.

verify.c
```diff
@@ -112,11 +112,11 @@ static inline size_t print_line(struct verifier_stream *stream, int eof)
 		return 0;
 
 	if (len > LOGLINELEN) {
-		fmt = stream->cut ? ".. %.*s .." : "%.*s ..";
+		fmt = stream->cut ? "verifier: .. %.*s .." : "%.*s ..";
```
wcawijngaards: This seems to be from a different issue. Code looks fine, by the way.

wcawijngaards (approved):
Looks like this would enable use of packaged versions, without needing a recompile. The new option is implemented with code in all the places it needs to.

dbaccess.c
wcawijngaards: Also in tpks/cutest/qtest.c:75 nsd->options->root_server=1; could be useful for the unit test. Or edit the files tpkg/cutest_qroot.tdir/root.conf root2.conf and unsigned.conf to set the option. If test modifications is a good solution here these places may also need modification.

nsd-checkzone.c
wcawijngaards: Perhaps a simple nsd->options->root_server = 1; could suffice here. The operator with a wrongly configured root zone can be told the configuration does not work from the nsd-checkconf utility, or when the server starts. The zone file itself, can be just fine, and when checked produce no error by itself.

anandb-ripencc:
I wrote in #275 that the `--enable-root-server` option is not necessary. NSD is the only one that has this restriction, but it does not make sense. It has been argued that it will prevent accidents, but I have never seen an accident happen with BIND, Knot DNS or PowerDNS, none of which have such a restriction. Can you please reconsider this, and DROP the entire option altogether? Make things simpler, not more complex, please.

anandb-ripencc:
If you can present an actual case of misconfiguration by an operator, leading to actual problems, I'd be willing to consider leaving this option in. Without such a case, and the mere *theoretical* danger foreseen by someone, is not enough to maintain this option.

anandb-ripencc:
Name servers like BIND are going down the path of less code, less complexity and leanness, and here we have an example of more code and more complexity. It really feels wrong.

k0ekk0ek:
Hi @anandb-ripencc. The option is to prevent people used to BIND from configuring a hints zone and thereby unintentionally serve a hint version of the root zone. Since this option has been there from the start, it'd make sense you did not see any issues pop up :slightly_smiling_face: For me personally, it's hard to state whether or not the option has actually prevented issues in the past or will help prevent them in the future. But, IMHO this is not the type of option that makes everything super complex or introduce bugs. Of course, your claim may still be valid. @wtoorop, @wcawijngaards any thoughts?

wcawijngaards:
I think that implementing the options that Anand likes, is a good idea, because it is a feature for a root server, and RIPE is a root operator. But in addition, one of the failures is loading a root.hints file in a root zone, something that is prevented by a parse failure due to a lack of a SOA record. It is trying to be idiot proof, and this is not really meant for Anand, but for mistakes by other users, but I have no examples of these mistakes; people porting BIND config files and adding root zones; but then Anand says those software configs do not have the problem, so maybe it is not a problem or no longer perceived as such a problem.

k0ekk0ek:
I didn't think of the "parse error" angle. If the file cannot be loaded to begin with, I think it makes sense to simply remove the option(s) and instead add a test (if it's not already there) to verify a hints file cannot be loaded(?)

wtoorop:
> Hi @anandb-ripencc. The option is to prevent people used to BIND from configuring a hints zone and thereby unintentionally serve a hint version of the root zone. Since this option has been there from the start, it'd make sense you did not see any issues pop up slightly_smiling_face For me personally, it's hard to state whether or not the option has actually prevented issues in the past or will help prevent them in the future. But, IMHO this is not the type of option that makes everything super complex or introduce bugs. Of course, your claim may still be valid. @wtoorop, @wcawijngaards any thoughts?

The change doesn't introduce any real or new complexity into the code, and it is the most conservative, most backwards compatible way to deal with the request (The default of the option is **yes** when `--enable-root-server` was used).  So I'd say the PR is the proper approach.

That said, I'm also just fine with only changing the default value of the `--enable-root-server` option in `configure.ac` as the only change to address this. But then we should give a small warning about the backwards compatible change in the release notes I guess.

anandb-ripencc:
I would like to maintain my view that the `--enable-root-server` option, as well as this new code, are all there for a THEORETICALLY perceived problem, one that I have never seen.

@wcawijngaards has already pointed out that the one possible scenario where users might get it wrong is the case of loading `root.hints`, but that won't work, as it is not a valid zone file. NSD will not load the zone, and there won't be an operational problem. I've been on the NSD mailing list for many years, and I've never heard of anyone asking why they could not load `root.hints`. Similarly, I've been on the Knot DNS mailing list for years, and have never come across anyone asking for this.

I am also unhappy about that idea that a new option in `nsd.conf`, called `root_server`, will default to either `on` or `off` depending on the compile time option. There will then be builds of NSD out there where the default behaviour will differ. One will only be able to tell the difference by looking at the nsd.conf man page. As a champion of consistency, I do not like this at all.

So I would still like to argue for dropping this `--enable-root-server` option altogether. You can emit a warning in case someone tries to use it.

anandb-ripencc:
@bjovereinder what do you think of this discussion? I know you're neither a developer, nor an end user, but do you see my point?

wtoorop:
For what it's worth, I do see your point @anandb-ripencc and I agree that the option would be confusing. So it looks to me we're all fine with dropping it (Jeroen was already suggesting that, and I also stated that I wouldn't mind).

bjovereinder:
Hi Anand,

> @bjovereinder what do you think of this discussion? I know you're neither a developer, nor an end user, but do you see my point?

Thanks.  Jeroen was already updating the PR that largely covered your concerns, with attention on backward compatibility.

-- Benno

k0ekk0ek:
Updated according to comments. @wcawijngaards, can you have a quick look to see if you agree with the changes?

wcawijngaards:
It looks okay to me.

PR #278, opened 2023-05-09
Replace Flex+Bison based zone parser with simdzone

k0ekk0ek:
There's still some warnings to fix and of course more tests to run. I only did very minor testing yet for NSD integration, but it's a good time to start the review as it's quite a bit (I think). This PR remove the Bison+Flex based zone parser from NSD and replaces it with [simdzone](https://github.com/NLnetLabs/simdzone).

Please note that both these changes and the current `main` branch of simdzone require review. This is also a good time to figure out how to best manage sub modules in the NSD repository. I now just did it commit based on the latest commit in master, we probably want to change that?


ixfr.c
```diff
@@ -2419,45 +2362,41 @@ static void clear_temp_table_of_rr(struct domain_table* temptable,
 
 /* read ixfr data new SOA */
 static int ixfr_data_readnewsoa(struct ixfr_data* data, struct zone* zone,
-	FILE* in, const char* ixfrfile, struct region* tempregion,
+	struct rr *rr, zone_parser_t *parser, struct region* tempregion,
```
wcawijngaards, line 2365: The `zone_parser_t`, the _t is owned by a standard, and not actually for application usage. We previously changed all of the type declarations in NSD for that reason, after that was reported. Perhaps 'zone_parser_type' or 'zone_parser' is a nice identifier.

ixfr.c
```diff
@@ -2512,76 +2447,163 @@ static int ixfr_data_readoldsoa(struct ixfr_data* data, struct zone* zone,
...
+	region_free_all(tempregion);
+	/* check SOA and also serial, because there could be other
+	 * add and del sections from older versions collated, we can
+	 * see this del section end when it has the serial */
+	if(rr->type == TYPE_SOA && soa_rr_get_serial(rr) == data->newserial) {
+		return 1;
 	}
 	ixfr_trim_capacity(&data->del, &data->del_len, &capacity);
-	return 1;
+	return 2;
```
wcawijngaards: The routine now returns 1 for type SOA. And 2 otherwise, but later on line 2561 it increases the soa_rr_count when '2' is returned. But that would increase the soa_rr_count for nonSOA types? Is that a mistake, should it be `!=` comparisons like in line 2484?

ixfr.c
```diff
@@ -2512,76 +2447,163 @@ static int ixfr_data_readoldsoa(struct ixfr_data* data, struct zone* zone,
...
+		zone_error(parser, "zone %s ixdr data: cannot allocate space",
+			zone->opts->name);
+		return 0;
+	}
+	clear_temp_table_of_rr(temptable, tempzone, rr);
+	region_free_all(tempregion);
+	/* check SOA and also serial, because there could be other
+	 * add and del sections from older versions collated, we can
+	 * see this del section end when it has the serial */
+	if(rr->type == TYPE_SOA && soa_rr_get_serial(rr) == data->newserial) {
```
wcawijngaards: In case `!=` is desired here, this is the suggestion
```suggestion
	if(rr->type != TYPE_SOA && soa_rr_get_serial(rr) != data->newserial) {
```

ixfr.c
```diff
@@ -2669,27 +2689,9 @@ static int ixfr_read_one_more_file(struct nsd* nsd, struct zone* zone,
 	const char* zfile, int num_files, uint32_t *dest_serial)
 {
 	char ixfrfile[1024+24];
-	FILE* in;
 	int file_num = num_files+1;
 	make_ixfr_name(ixfrfile, sizeof(ixfrfile), zfile, file_num);
-	in = fopen(ixfrfile, "r");
-	if(!in) {
-		if(errno == ENOENT) {
```
wcawijngaards, line 2677: The test for end of ixfr number sequence is removed, but instead there is the zone_parse call. Does that log errors, because errors have to be logged somehow? Is the code path to deal with the file after the last existing ixfr file okay, and does it then terminate without overly logging errors for that. Perhaps fopen and close the file to test for ENOENT en also the warn_if_directory call, that is there for usability for operator configuration. The return value of zone_parse is otherwise correct, the 0 return value would make things work, this is a note about logs and checks.

zonec.c
```diff
@@ -2190,183 +150,256 @@ check_dname(zone_type* zone)
...
+
+	dname = dname_make(state->database->region, owner->octets, 1);
+	assert(dname);
+	domain = domain_table_insert(state->domains, dname);
+	assert(domain);
+
+	rdata_count = rdata_wireformat_to_rdata_atoms(
+		state->database->region, state->domains, type, rdlength, &buffer, &rdatas);
+	// rdlength must not exceed maximum of 65535 bytes
+	assert(rdlength < MAX_RDLENGTH);
```
wcawijngaards: Is there a reason why 65535 itself is not allowed? Perhaps use <= for the comparison, but that is then always true due to the integer type.

zonec.c
```diff
@@ -2190,183 +150,256 @@ check_dname(zone_type* zone)
...
+	// number of atoms must not exceed maximum of 65535 (all empty strings)
+	assert(rdata_count >= 0);
+	assert(rdata_count < MAX_RDLENGTH);
```
wcawijngaards: Is there a reason why 65535 itself is not allowed? Perhaps use <= for the comparison, but that is then always true due to the integer type.

zonec.c
```diff
@@ -2190,183 +150,256 @@ check_dname(zone_type* zone)
...
+
+	/* we have the zone already */
+	if (type == TYPE_SOA) {
+		if (domain != state->zone->apex)
+			zone_log(parser, priority, "SOA record with invalid domain name, '%s' is not '%s'", "x", "y");
```
wcawijngaards: Placeholder "x" and "y" are here instead of the strings.

zonec.c
```diff
@@ -2190,183 +150,256 @@ check_dname(zone_type* zone)
...
+	struct zonec_state *state = (struct zonec_state *)user_data;
+
+	assert(state);
+
+	struct rr *rr;
+	struct rrset *rrset;
+	const struct dname *dname;
+	struct domain *domain;
+	struct buffer buffer;
+	int priority = ZONE_ERROR << (parser->options.non_strict == true);
```
wcawijngaards: So this compares the boolean with a 'true' value, uses the numerical value of the comparison, then uses that shifts define the ZONE_ERROR and ZONE_WARNING priorities in zone.h. It could also be with '?', or is that to stop alternate code paths, for optimization? In which case that could be annotated; that non_strict has to have 'true', not unzero, and the shifts are used in this manner. Maybe use '1' instead of 'true'?

zonec.c
```diff
@@ -2190,183 +150,256 @@ check_dname(zone_type* zone)
...
+	}
+
+	/* Check if zone file contained a correct SOA record */
+	if (!zone) {
+		log_msg(LOG_ERR, "zone configured as '%s' has no content.", name);
+		state.errors++;
+	} else if (!zone->soa_rrset || zone->soa_rrset->rr_count == 0) {
+		log_msg(LOG_ERR, "zone configured as '%s' has no SOA record", name);
+		state.errors++;
+#if 0
```
wcawijngaards: This check is not disabled in the original zonec_read function.

zonec.c
```diff
@@ -2190,183 +150,256 @@ check_dname(zone_type* zone)
...
-	else	*parsed = parser->prev_dname;
-	/* remove origin if it was not used during the parse */
-	if(parser->origin != error_domain)
-		domain_table_deldomain(parser->db, parser->origin);
-	region_free_all(parser->rr_region);
-	zonec_desetup_string_parser();
-	parser_flush();
-	return errors;
-}
+	startzonec = time(NULL);
```
wcawijngaards: This time seems to be unused.

wcawijngaards (approved):
The code looks nice! There are some small notes. The submodule was causing ci to fail, also makedist.sh needs to include submodules in the source tarball output.

ixfr.c
k0ekk0ek: In NSD it's a problem because no prefixing is done. simdzone uses the prefix `zone_` consistently for all exported symbols. So, while `_t` is technically reserved by the specification, I'd say chances of name conflicts are very very minimal, especially because it's common practice to use `_t`. The case is very different from using names starting with `_` or `__`.

ixfr.c
k0ekk0ek: Indeed. Thanks :+1:

ixfr.c
k0ekk0ek: Fixed.

ixfr.c
k0ekk0ek: Yes, the parser logs errors. The idea is that the application registers a log callback function of type `zone_log_t` in `options.log.callback` that gets called if an error occurs. The function can then format/decorate the error as desired and pass it to it's actual log function. i.e. `log_msg` in the case of NSD. While in the callback, the application can use `zone_log` (or `zone_error`, `zone_warning` or `zone_info`) to log a message using the callback registered by the application. That's how it's supposed to work at least. In this case the parser should print ab error message and return `ZONE_READ_ERROR` (or `ZONE_OUT_OF_MEMORY` in case of `ENOMEM`). I did forget about implementing the printing a message part though :sweat_smile: (so, I'll fix this in an upcoming commit)

zonec.c
k0ekk0ek: Actually, `rdata_count` is an `ssize_t` (unlike comment above), so the limit is higher. Thanks for spotting my wrong assumption!

zonec.c
k0ekk0ek: Removed this check. Thanks for pointing out the wrong assumption :+1:

zonec.c
k0ekk0ek: Good catch

zonec.c
k0ekk0ek: Removed

zonec.c
k0ekk0ek: Correct. I disabled it for now. I'll have to turn the owner into a `dname` and then compare. I'll fix in an upcoming commit.

zonec.c
k0ekk0ek: Yeah, it's probably better to change this for readability. Will fix.

ixfr.c
wcawijngaards: But the test for ENOENT is necessary for the routine that calls the read of ixfr files. Hence that may need more attention and not just deleting the check, as it goes from a check on ENOENT to a version of the code without a check for ENOENT, and I think that likely means the case may not be handled as it should. I could be nice to handle it by itself anyway for the reason that it is part of the read of ixfr files routine.

zonec.c
wcawijngaards: The comment is on the 'rdlength' line. But nice that the also included line on rdata_count can be improved!

wcawijngaards: But that is another comment, for a similar issue.

zonec.c
wcawijngaards: The fix prints `state->zone->apex` twice, but not `domain`.

k0ekk0ek:
Rebased on current `master` and fixed `soa_test.tdir/soa_test.zone5` (no previous owner). Only test failing is `bug137_compile_outdir.tdir` which has to do with compilation, not actual functionality.

zonec.c
k0ekk0ek: Fixed

ixfr.c
k0ekk0ek: This is fixed by NLnetLabs/simdzone#185.

wcawijngaards: Apart from the parser making an error callback, the nsd side of the code has to actually handle the file nonexisting case. Because it has to detect that the ixfr file does not exist, at this place or somewhere else in the NSD code that deals with the ixfr files.

k0ekk0ek: Can you elaborate? I think I'm overlooking a case(?) `ixfr_read_from_file` repeatedly calls `ixfr_read_one_more_file` until there's no more files. Judging by the code that was previously there, no distinction is made between a file not being there or an error condition(?) Therefore, the operation stays the same? At least, that was the intent. The zone parser reads the file, if it fails, `0` (zero) is returned, otherwise `1`.

zonec.c
k0ekk0ek: I've re-enabled this check btw. Comparing the owner to the origin, which is constructed from the zone name.

ixfr.c
wcawijngaards: There is a replacement of the old lines 2677-2681 with what is now 2664, that calls 2598, that is really like old lines 2682-2683. In references:
https://github.com/NLnetLabs/nsd/pull/278/files#diff-c2dbc9ea381942e85b224edd25e15642084afe38b89872a7e14767760c399476#L2677-L2681
turns into
https://github.com/NLnetLabs/nsd/pull/278/files#diff-c2dbc9ea381942e85b224edd25e15642084afe38b89872a7e14767760c399476R2664
and
https://github.com/NLnetLabs/nsd/pull/278/files#diff-c2dbc9ea381942e85b224edd25e15642084afe38b89872a7e14767760c399476R2598
but that is like
https://github.com/NLnetLabs/nsd/pull/278/files#diff-c2dbc9ea381942e85b224edd25e15642084afe38b89872a7e14767760c399476L2682-L2683

wcawijngaards: The code links did not look at pretty as I wanted. So the error code that is returned is fine, and now the parse is fixed it always prints an error. In the old sequence of code, it would not print an error, because it is not an error to find the final ixfr file, and then stop processing the list of ixfr files. Does the new code now print an error every time a list of ixfr files is read?

k0ekk0ek: Thanks, I think I get what you're saying. If the file that is passed to `zone_parse` cannot be found, it is considered an error by the parser. In this case, that should not happen because it's expected behavior. To fix that, I can check if the `file` is `NULL` in `ixfr_data_log` and then simply not print the message. `ixfr_data_read` would then return `0`, which would lead to `ixfr_read_from_file` stopping the loop. I think, with the last addition of not printing the message, the behavior will be the same as before?

wcawijngaards: Yes, I think that would work! The code could also call stat() on the file before opening it, is that is more convenient.

k0ekk0ek: I've chosen to add a call to `stat`. Thanks :+1:

k0ekk0ek:
This PR builds on all platforms, except for a warning that is not related to these changes(?) At least, the link escapes me for now. I'll take this one tomorrow.
```
gcc-11 -I. -Isimdzone/include -g -O2 -W -Wall -Wextra -Wdeclaration-after-statement -c xfrd-catalog-zones.c
In function 'xfr_writer_init',
    inlined from 'xfrd_process_catalog_producer_zone' at xfrd-catalog-zones.c:1190:2:
xfrd-catalog-zones.c:1114:9: warning: 'xw' may be used uninitialized [-Wmaybe-uninitialized]
 1114 |         buffer_create_from( &xw->packet, &xw->packet_space
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1115 |                                        , sizeof(xw->packet_space));
      |                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from dname.h:16,
                 from namedb.h:15,
                 from difffile.h:13,
                 from xfrd-catalog-zones.c:9:
xfrd-catalog-zones.c: In function 'xfrd_process_catalog_producer_zone':
buffer.h:79:6: note: by argument 2 of type 'const void *' to 'buffer_create_from' declared here
   79 | void buffer_create_from(buffer_type *buffer, const void *data, size_t size);
      |      ^~~~~~~~~~~~~~~~~~
xfrd-catalog-zones.c:1182:32: note: 'xw' declared here
 1182 |         struct xfrd_xfr_writer xw;
      |                                ^~
```

zonec.c
```diff
@@ -265,7 +265,7 @@ int32_t zonec_accept(
 		struct rr *rrs;
 		if (type != TYPE_RRSIG && ttl != rrset->rrs[0].ttl) {
 			zone_log(parser, priority, "%s TTL %"PRIu16" does not match TTL %u of %s RRset",
-				domain_to_string(domain), ttl, rrset->rrs[0].ttl,
+				domain_to_string(domain), (uint16_t)ttl, rrset->rrs[0].ttl,
```
wcawijngaards: The ttl variable is uint32_t, not uint16_t.

k0ekk0ek: Indeed. Format string was incorrect. Fixed.

k0ekk0ek:
All checks pass. I'll rebase and squash the debug commits so the PR can be merged.

k0ekk0ek:
I've consolidated all commits. Tests are all passing, review comments have all been addressed. I think this can be merged.

PR #281, opened 2023-05-23
Proxy protocol

wcawijngaards:
This is an implementation of PROXYv2 for NSD.

It can be configured with `proxy-protocol-port: <portnum>` with the port number of the interface on which proxy traffic is handled. The interface can support proxy traffic for UDP, TCP and TLS.

It uses code from Unbound, that implements the proxy protocol specs, in https://github.com/NLnetLabs/unbound/pull/881 .


server.c
```diff
@@ -131,6 +132,16 @@ log_addr(const char* descr,
   #define TCP_FASTOPEN_SERVER_BIT_MASK 0x2
 #endif
 
+/* header state for the PROXYv2 header (for TCP) */
+enum pp2_header_state_type {
```
k0ekk0ek: Normally, we reserve `_type` for `typedef`'s. Maybe `typedef enum pp2_header_state { ...} pp2_header_state_type` is more consistent?

server.c
```diff
@@ -131,6 +132,16 @@ log_addr(const char* descr,
...
+	/* no header encounter yet */
+	pp2_header_none = 0,
+	/* read the static part of the header */
+	pp2_header_init,
+	/* read the full header */
+	pp2_header_done
```
k0ekk0ek, line 142: Perhaps it's better to use uppercase for `enum` labels instead?

server.c
```diff
@@ -3570,6 +3603,88 @@ port_is_zero(
...
+		VERBOSITY(4, (LOG_ERR, "proxy-protocol: not enough buffer "
+			"size to read PROXYv2 header"));
+		return 0;
+	}
+	if((header->ver_cmd & 0xF) == PP2_CMD_LOCAL) {
+		/* A connection from the proxy itself.
+		 * No need to do anything with addresses. */
+		goto done;
+	}
+	if(header->fam_prot == 0x00) {
```
k0ekk0ek: Why do this check outside the `switch`?

options.c
```diff
@@ -2578,3 +2630,32 @@ resolve_interface_names(struct nsd_options* options)
...
+	struct sockaddr* addr)
+{
+	struct proxy_protocol_port_list* p;
+	int port;
+#ifdef INET6
+	struct sockaddr_storage* ss = (struct sockaddr_storage*)addr;
+	if(ss->ss_family == AF_INET6) {
+		struct sockaddr_in6* a6 = (struct sockaddr_in6*)addr;
+		port = ntohs(a6->sin6_port);
+	} else if(ss->ss_family == AF_INET) {
```
k0ekk0ek, line 2645: This seems incorrect? As in, protocol can still be something else than AF_INET if INET6 isn't defined.

util/proxy_protocol.c
```diff
@@ -0,0 +1,194 @@
...
+	if(buflen < expected_size) {
+		return 0;
+	}
+	/* sig */
+	memcpy(buf, PP2_SIG, PP2_SIG_LEN);
+	buf += PP2_SIG_LEN;
+	/* version and command */
+	*buf = (PP2_VERSION << 4) | PP2_CMD_PROXY;
+	buf++;
+	if(af==AF_INET) {
```
k0ekk0ek: This check is inconsistent? We define `INET`, `INET6` and `UNIX` socket types, but we only support `INET` and `INET6` here?

util/proxy_protocol.c
```diff
@@ -0,0 +1,194 @@
...
+		return PP_PARSE_SIZE;
+	}
+	/* Check for supported commands */
+	if((header->ver_cmd & 0xF) != PP2_CMD_LOCAL &&
+		(header->ver_cmd & 0xF) != PP2_CMD_PROXY) {
+		return PP_PARSE_UNKNOWN_CMD;
+	}
+	/* Check for supported family and protocol */
+	if(header->fam_prot != 0x00 /* AF_UNSPEC|UNSPEC */ &&
+		header->fam_prot != 0x11 /* AF_INET|STREAM */ &&
```
k0ekk0ek: Maybe it's better to define constants for the combinations instead? So for instance `PP2_INET_STREAM` etc? Makes it easier to read.

util/proxy_protocol.h
```diff
@@ -0,0 +1,164 @@
...
+ */
+enum pp2_command {
+	PP2_CMD_LOCAL = 0x0,
+	PP2_CMD_PROXY = 0x1
+};
+
+/**
+ * PROXYv2 address family.
+ */
+enum pp2_af {
```
k0ekk0ek, line 68: Maybe specify these definitions originate from the PROXYv2 protocol specification and are passed on the wire?

server.c
```diff
@@ -3854,10 +3986,136 @@ handle_tcp_reading(int fd, short event, void* arg)
...
+				cleanup_tcp_handler(data);
+				return;
+			}
+			VERBOSITY(6, (LOG_INFO, "proxy-protocol: reading variable part of PROXYv2 header (len %lu)", (unsigned long)want_read_size));
+			current_read_size = PP2_HEADER_SIZE + want_read_size;
+			if(want_read_size == 0) {
+				/* nothing more to read; header is complete */
+				data->pp2_header_state = pp2_header_done;
+			} else if(data->bytes_transmitted < current_read_size) {
+				received = read(fd,
```
k0ekk0ek: Seems we could benefit from implementing a function to read+handle errors instead of repeating the same? Can be postponed of course, but seems convenient

k0ekk0ek (approved):
Some small remarks about things that stood out. Nothing blocking.

server.c
wcawijngaards: Fixed in https://github.com/NLnetLabs/nsd/pull/281/commits/427739ac972b9395ba5d257e381115ce9ddc19b2 . Used the type name without the `_type` suffix.

server.c
wcawijngaards: Because this is not a set of type identifiers, like the enum for like the RR type list is, I think the lowercase is fine. It is also in other places like that. So, I do not see why to change the code style. Thanks for the heads up, though.

server.c
wcawijngaards: This is also in the Unbound proxy protocol code. I think it is because the 0x00 check is for an out of band message, of a healthcheck type, and it does not contain an address. In the switch it deals with the address, so this skips the address management part, to do other commands.

Apart from that, the code could be in the switch, technically. So I would not mind really, either way. I'll keep it like it is now, because of the flow of the algorithm from the comments.

options.c
wcawijngaards: Fixed in https://github.com/NLnetLabs/nsd/pull/281/commits/944aa2bdef5fac1d9b68de800c0f6cb7da1cd4c5 . That checks for the sockaddr family also when INET6 is not defined. The code does not actually have other sockets than AF_INET and AF_INET6 sockets for this part, so this is a depth defense check, that is not really needed, like an assert.

util/proxy_protocol.c
wcawijngaards: This code, util/proxy_protocol.c is from the generic proxy protocol code, that is in Unbound.

The code really only has `INET` and `INET6` socket types that use these routines, so it is safe. But from a defense in depth point, it is possible to have the extra checks. That said, I am fine with either change, of an added check or not.

util/proxy_protocol.c
wcawijngaards: This code, util/proxy_protocol.c is from the generic proxy protocol code, that is in Unbound.

The constants for it as defined in util/proxy_protocol.h. I think it was chosen to display like this to be more readable, than the bit munging that is needed with the constants, because the values align with hex characters. That said, I am fine with this code but also with a use of the defined constants.

util/proxy_protocol.h
wcawijngaards: This code, util/proxy_protocol.h is from the generic proxy protocol code, that is in Unbound.

You think that would improve the comment, I would have thought that fairly obvious. But that said, I am okay with the change. More comments can make for more readable and hence better maintained code.

server.c
wcawijngaards: It is in commit https://github.com/NLnetLabs/nsd/pull/281/commits/db581ac35ed36ee59694e967e313f949da676862 . That creates functions for TCP and TLS to read more data and handle an error.

pettai:
Hi,
Any update on when this is going to be merged into the main branch?

k0ekk0ek:
Hi @pettai, it's waiting for one more review to complete. Expect this to be merged soon.

pettai:
I noted an issue #297 then merging this branch to the latest git version of NSD, so apart from the review, this also needs to be fixed before merging. Thx

wcawijngaards:
The commit https://github.com/NLnetLabs/nsd/commit/dfcdadb828f6484d82c7219a2990cd142d716885 fixes the issue. The buffer was not reset properly between queries, the commit fixes that.

util/proxy_protocol.c
gthess: Since this code is supposed to be generic I agree with @k0ekk0ek 's observation although the code is not used that way (yet). Turning this into a switch statement where `UNIX` and `default` would return 0 seems like a nice approach.

util/proxy_protocol.h
gthess: I see what @k0ekk0ek means and it would help differentiate between these enums and `pp_parse_errors` further down for example. Would something like `PROXYv2 address family (wire value)` work for this comment and the rest here?

util/proxy_protocol.c
gthess: Defined constants should work better since I see that these combinations are also used elsewhere in the code.

gthess (approved):
The PR looks nice!
Just replied on @k0ekk0ek 's suggestions.

util/proxy_protocol.c
gthess: Resolved in d8a47b171ca92092af8c33a888487726f72c77b4.

util/proxy_protocol.c
gthess: Resolved in d8a47b171ca92092af8c33a888487726f72c77b4.

util/proxy_protocol.h
gthess: Resolved in d8a47b171ca92092af8c33a888487726f72c77b4. Used `(protocol value)` instead.

pettai:
So it has been done 🎉

PR #282, opened 2023-06-03
Improve nsd.conf man page

anandb-ripencc:
This commit fixes some small typos, adds punctuation for readability, and brings more consistency to some text.

### DIFF
--- a/nsd.conf.5.in
+++ b/nsd.conf.5.in
@@ -99,7 +99,7 @@ There must be whitespace between keywords. Attribute keywords end
 with a colon ':'. An attribute is followed by its containing 
 attributes, or a value. 
 .P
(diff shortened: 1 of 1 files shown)


anandb-ripencc:
@wcawijngaards would be nice if you merge this before releasing 4.7.0.

wcawijngaards:
It is in a freeze period, because the RC1 is already out for 4.7.0. So this type of change is better for after that.

Looks like a good couple of fixups, anyhow.

wcawijngaards (approved):
Small text errors to improve on the man page, they are good to have.

wcawijngaards:
Thanks for the fixes! They are added in the code repository.

PR #291, opened 2023-09-07
Check for __b64_pton and __b64_ntop existence and declaration.

millert:
If present, and the non-underscore versions don't exist, define to b64_pton and b64_ntop.  These functions are usually defined in resolv.h, which nsd does not include, so use the prototype in config.h if needed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -849,8 +849,12 @@ AC_REPLACE_FUNCS(snprintf)
 AC_REPLACE_FUNCS(strlcat)
 AC_REPLACE_FUNCS(strlcpy)
 AC_REPLACE_FUNCS(strptime)
-AC_REPLACE_FUNCS(b64_pton)
-AC_REPLACE_FUNCS(b64_ntop)
+AC_CHECK_FUNCS(b64_pton, [AC_CHECK_DECLS(b64_pton)], [
+  AC_CHECK_FUNCS(__b64_pton, [AC_CHECK_DECLS(__b64_pton)], [AC_LIBOBJ(b64_pton)])
+])
+AC_CHECK_FUNCS(b64_ntop, [AC_CHECK_DECLS(b64_ntop)], [
+  AC_CHECK_FUNCS(__b64_ntop, [AC_CHECK_DECLS(__b64_ntop)], [AC_LIBOBJ(b64_ntop)])
+])
 AC_REPLACE_FUNCS(pselect)
 AC_REPLACE_FUNCS(memmove)
 AC_REPLACE_FUNCS(setproctitle)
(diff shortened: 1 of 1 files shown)


k0ekk0ek:
Hi @millert! Thanks for the PR. Seems `resolv.h`, at least on Linux, defines `b64_xtox` to `__b64_xtox` too, so mostly looks fine to me. Can you elaborate on what system doesn't have that define? That sort of info helps to understand why this is good to merge. Also, you use `AC_CHECK_DECLS` to test for the `b64_xtox` symbols, but as `b64_xtox` is always defined to the underscore version, that check must always fail right?

For my own reference (maybe you're interested too, it states why `resolv.h` isn't included):
It seems someone else tried to improve the situation as well: https://lists.nlnetlabs.nl/pipermail/nsd-users/2015-September/002174.html

I'll have to read a bit more before I can determine if this is a correct fix.

millert:
I don't know of any systems that have the non-underscore versions that do not also have the underscore version.  For instance, Solaris has both (one is a weak ref) in its libresolv but not in libc.  Only checking for the underscore version would simplify configure a bit.

k0ekk0ek:
@millert, what system are you using that is currently having problems building NSD (so that I can test and verify)? And, are you experiencing the same problem that's outlined in the post? i.e. are the versions in `compat` being compiled and used as opposed to the implementation on your system (The post is from 2015 and that part of `configure.ac` hasn't been updated since 2013)?

The old post in nsd-users suggests using the underscore versions, but simply updating the check to actually compile code might work(?)

FYI: It also just so happens that we're in the process of creating a better zone parser all together ([simdzone](https://github.com/NLnetLabs/simdzone)).

k0ekk0ek:
Hi @millert. Finally got around to this PR. The `b64_*` variants in `compat` are not the best and we should swap them out eventually, but unfortunately we cannot use the functions that come with the operating system.

We don't want to depend on libresolv, because we don't want to have DNS client code in NSD. NSD must not assume the DNS is operational as it is an authoritative server (which is also why it insists on ip-addresses in it's configuration). While that's not a problem on BSD because these functions are in libc, we'd still need to include `resolv.h` and that (potentially) causes naming conflicts. We could declare the function ourselves, but that isn't a great idea either as the signature might change (not likely, but still).

All-in-all, we agreed it's probably best to just replace the implementations in `compat` with better ones and make them non-optional.

I'm sorry it took this long for us to come to this conclusion, I wasn't aware of the reasoning myself :sweat_smile:.


I did update the patch to work on Linux and check `resolv.h`, etc. We won't be using it, but I might as well post my comments/updates:

```m4
define([INCLUDES_BASE64],[
AC_INCLUDES_DEFAULT
#ifdef HAVE_NETINET_IN_H
# include <netinet/in.h>
#endif
#ifdef HAVE_RESOLV_H
# include <resolv.h>
#endif
])

AC_CHECK_HEADERS([netinet/in.h]) # needed to check for resolv.h on FreeBSD
AC_CHECK_HEADERS([resolv.h],,,[INCLUDES_BASE64])


# b64_pton and b64_ntop are typically defined to __b64_pton and __b64_ntop.
# b64_* symbols are not likely to exist, search for __b64_* symbols if that
# is the case. __b64_* symbols may reside in libresolv (Linux) or libc (BSDs).
AC_SEARCH_LIBS([b64_ntop],[resolv],[
  AC_CHECK_DECLS([b64_ntop],,,[INCLUDES_BASE64])],[
  AC_SEARCH_LIBS([__b64_ntop],[resolv],[
    AC_CHECK_DECLS([__b64_ntop],,,[INCLUDES_BASE64])],[AC_LIBOBJ(b64_ntop)])
])

AC_SEARCH_LIBS([b64_pton],[resolv],[
  AC_CHECK_DECLS([b64_pton],,,[INCLUDES_BASE64])],[
  AC_SEARCH_LIBS([__b64_pton],[resolv],[
    AC_CHECK_DECLS([__b64_pton],,,[INCLUDES_BASE64])],[AC_LIBOBJ(b64_pton)])
])

AC_SUBST([HAVE_B64_PTON])
AC_SUBST([HAVE_B64_NTOP])
AC_SUBST([HAVE_DECL_B64_PTON])
AC_SUBST([HAVE_DECL___B64_PTON])
AC_SUBST([HAVE_DECL_B64_NTOP])
AC_SUBST([HAVE_DECL___B64_NTOP])

AH_BOTTOM([
#ifdef HAVE_NETINET_IN_H
# include <netinet/in.h>
#endif
#ifdef HAVE_RESOLV_H
# include <resolv.h>
#endif

#if !HAVE_DECL_B64_NTOP
# if HAVE_DECL___B64_NTOP
#   define b64_ntop __b64_ntop
# else
int b64_ntop(uint8_t const *src, size_t srclength,
	     char *target, size_t targsize);
# endif
#endif

#if !HAVE_DECL_B64_PTON
# if HAVE_DECL___B64_PTON
#   define b64_pton __b64_pton
# else
int b64_pton(char const *src, uint8_t *target, size_t targsize);
# endif
#endif
])
```

PR #292, opened 2023-09-08
Faster stats

### LINKED ISSUE #286: NSD not responding to stats_noreset command while writing zone to file
Hi,

we recently stumbled upon an issue with NSD v4.6.1 in combination with the [nsd_exporter](https://github.com/optix2000/nsd_exporter), where the latter appeared to be unable to get any metrics from NSD at certain times.

The nsd_exporter uses the `stats_noreset` command to fetch the data from NSD and then does $THINGS to represent them in prometheus-readable format (it uses [go-nsdctl](https://github.com/optix2000/go-nsdctl) for communicating with NSD, which is basically a golang version of your very own nsd-control).

We were able to correlate the timing of this issue to the times where NSD is writing a zone file to disk (as specified in our `nsd.conf` with `zonefiles-write`), which takes quite a bit of time (think 60+ seconds, it's a rather large zone + slow disk).

During this time, NSD _does_ accept incoming connections from the exporter on the control channel (sends a TCP SYN back), but it does not send any data until the write operation is complete (but the connection is kept open).

Once the write is complete, all the outstanding commands in the still open connections are executed / answered (and the connections closed afterwards).

Since the write is taking so long, the client attempts multiple times to retrieve data from NSD while it is still writing the zone to disk, so at some point NSD refuses the new connections due to its built-in [maximum number of control connections](https://github.com/NLnetLabs/nsd/blob/3ad1ec071719f77e26b880f902d564bc88a3f496/remote.c#L612).

I understand that this is probably a weird edge-case (because honestly, who expects the write to take _that_ long :smile: ), but I would have expected the write operation not to "block" the collection of zone stats.

We also ran `nsd-control status`, `nsd-control zonesstatus` and `nsd-control stats_noreset` while the zone file was being written - the two status commands work just fine, only the `stats_noreset` hangs.

Unfortunately my knowledge of C is basically nonexistent so I am not sure where to even start looking how this might be fixed / changed...

Any feedback would be greatly appreciated!

wcawijngaards:
Process the nsd-control stats faster. It uses shared memory to convey the statistics from the server processes to the xfrd process. This is faster, and also works while a reload is in progress. The statistics are no longer written over the command pipes between processes, and so do not wait for the processes. It is similar to how zone-stats have been implemented. It works for both stats and stats_noreset.

Fix #286 .


server.c
```diff
@@ -687,6 +687,65 @@ server_zonestat_switch(struct nsd* nsd)
...
+void
+server_stat_alloc(struct nsd* nsd)
+{
+	char tmpfile[256];
+	size_t sz = sizeof(struct nsdst) * nsd->child_count * 2;
+	uint8_t z = 0;
+
+	/* file name */
+	nsd->statfname = 0;
+	snprintf(tmpfile, sizeof(tmpfile), "%snsd-xfr-%d/nsd.%u.stat",
```
k0ekk0ek, line 700: Maybe use a different filename pattern? `%snsd-xfr-%d/nsd.%u.stat` may cause confusion (the `-xfr-` part)? Don't know if this is on purpose (tooling may depend on it), so just a very minor remark.

server.c
```diff
@@ -687,6 +687,65 @@ server_zonestat_switch(struct nsd* nsd)
...
+		exit(1);
+	}
+
+#ifdef HAVE_MMAP
+	if(lseek(nsd->statfd, (off_t)sz-1, SEEK_SET) == -1) {
+		log_msg(LOG_ERR, "lseek %s: %s", nsd->statfname,
+			strerror(errno));
+		unlink(nsd->statfname);
+		unlink(nsd->zonestatfname[0]);
+		unlink(nsd->zonestatfname[1]);
```
k0ekk0ek: Maybe use a single error handler? Each failure case has the three exact same operations and then `exit(1)`. Maybe use a `goto` or function? Also, isn't it the case that if you unlink, but keep the file descriptor open, the file is gone from the file system, but the file descriptor remains valid? In that case, technically, a `close` should be done first (not that it matters after `exit(1)`).

server.c
```diff
@@ -2319,6 +2378,10 @@ server_reload(struct nsd *nsd, region_type* server_region, netio_type* netio,
 	/* Restart dumping stats if required.  */
 	time(&nsd->st->boot);
 	set_bind8_alarm(nsd);
+	/* Switch to a different set of stat array for new server processes,
+	 * because they can briefly coexist with the old processes. They
+	 * have their own stat structure. */
+	nsd->stat_current = (nsd->stat_current==0?1:0);
```
k0ekk0ek, line 2347: You could write this as: `!nsd->stat_current`(?)

server.c
```diff
@@ -2813,6 +2847,9 @@ server_main(struct nsd *nsd)
 	unlink(nsd->zonestatfname[0]);
 	unlink(nsd->zonestatfname[1]);
 #endif
+#ifdef BIND8_STATS
+	unlink(nsd->statfname);
```
k0ekk0ek: Maybe we should add a `server_stat_free` to handle cleanup? The inverse of `server_stat_alloc`, may make reasoning about these operations easier.

k0ekk0ek (approved):
Some minor remarks that are more-or-less personal preferences. Overall this looks good, nice reduction in complexity with regard to ipc and better usability :slightly_smiling_face:

server.c
wcawijngaards: The same directory is also used in server_zonestat_alloc server.c:540, and server_prepare_xfrd server.c:1622. For the zone stats and the difffile task mmap files. So I thought to put them in the same place. This directory is sitting in a temp directory and the zone transfers are also there. Since the directory can get the correct access permissions putting the mmapped file in there is a good idea I think.

The name of the directory stems from the use of xfrd to store zone transfers in it. Perhaps we could rename it, it is dynamically created and removed when nsd starts.

k0ekk0ek: I figured something along those lines. Feel free to keep it as-is :+1:

server.c
wcawijngaards: Fixed in https://github.com/NLnetLabs/nsd/pull/292/commits/9557fd283e1a748f206ec534fa0efbd3774eb5d5 . This closes the statfd.

server.c
wcawijngaards: I think that is true, but may depend on compiler version and language specification. So I prefer to be explicit about it here.

server.c
wcawijngaards: The `server_stat_free` function is added in this commit https://github.com/NLnetLabs/nsd/pull/292/commits/9b4090bc912fd9933bf6480338c9c93ee49d8dad . Maybe that is easier to deal with.

PR #294, opened 2023-09-19
Catalog Zones

Koenvh1:
This PR implements catalog zone consumers as defined in [RFC 9432](https://datatracker.ietf.org/doc/html/rfc9432). This is a squash and rebase of #261 onto a recent version of the master branch to remove most of the noise caused by earlier implementation ideas without rewriting history on the previous branch.

EDIT (by @k0ekk0ek): `features/catalog-zones` is a feature branch that will collect the changes by @Koenvh1 (primary author) and @k0ekk0ek.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -81,12 +81,12 @@ MANUALS=nsd.8 nsd-checkconf.8 nsd-checkzone.8 nsd-control.8 nsd.conf.5
 
 COMMON_OBJ=answer.o axfr.o ixfr.o ixfrcreate.o buffer.o configlexer.o configparser.o dname.o dns.o edns.o iterated_hash.o lookup3.o namedb.o nsec3.o options.o packet.o query.o rbtree.o radtree.o rdata.o region-allocator.o rrl.o siphash.o tsig.o tsig-openssl.o udb.o udbradtree.o udbzone.o util.o bitset.o popen3.o
 XFRD_OBJ=xfrd-disk.o xfrd-notify.o xfrd-tcp.o xfrd.o remote.o $(DNSTAP_OBJ)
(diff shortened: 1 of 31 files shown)


pettai:
A noted behavior-change is that the catalog zone is never written to disk. Is that change from the previous branch intentional?

k0ekk0ek:
@pettai, If a database is configured, the zone (on secondaries) is not written to disk. If no database is configured `database: ""` zones should be written to disk every 3600 seconds by default (configurable, see `zonefiles-write`). Note that the database option is set to be removed in the next release (#290). It could (of course) also be that there's a bug introduced by this PR. In the latter case, I'll happily dive into it, but please note that at this stage, the code is not quite ready for production testing. Reviews of #261 turned up some problems that @Koenvh1 and I are working to resolve. We're hoping to have something ready in a few weeks.

pettai:
> @pettai, If a database is configured, the zone (on secondaries) is not written to disk. If no database is configured `database: ""` zones should be written to disk every 3600 seconds by default (configurable, see `zonefiles-write`). Note that the database option is set to be removed in the next release (#290). It could (of course) also be that there's a bug introduced by this PR. In the latter case, I'll happily dive into it, but please note that at this stage, the code is not quite ready for production testing. Reviews of #261 turned up some problems that @Koenvh1 and I are working to resolve. We're hoping to have something ready in a few weeks.

I just used `nsd-control write` before on the previous catz conf to write all zones in memory (after transfer) to disk. But after adding a explicit zonefile: statement to the catalog zone: itself, it is now written to disk 👍 
The new configuration option(s) follows nsd's generic behavior, so I rest my case 😄

k0ekk0ek:
New implementation in #304. Closing this.

PR #296, opened 2023-09-21
Make catalog zone settings local to zone

k0ekk0ek:
This PR moves catalog zone specific settings out of the pattern. I've also included a test that ensures catalog produces do not magically instantiate zones that are in the catalog zone but not configured. I've also added that queries to catalog zones are disallowed by default (still need to add a test for that). @Koenvh1, @wtoorop, do we agree the producer should not instantiate zones based on the catalog zone?

### DIFF
--- a/cat-zones.c
+++ b/cat-zones.c
@@ -83,6 +83,8 @@ void catalog_consumer_process(
 	catzonezone_type* catzonezones = NULL;
 	region_type* catzonezones_region = region_create(xalloc, free);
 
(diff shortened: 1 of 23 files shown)


k0ekk0ek:
Marked as draft. After some discussions it turns out we do want to instantiate zones on the primary, but what exactly is done in that case differs from what happens on the secondary. In other words: this still needs some work.

k0ekk0ek:
Don't need this anymore as catalog zones is already in.

PR #300, opened 2023-10-23
Fix IXFR-out after intermediate update fails verification

k0ekk0ek:
This PR is work in progress! Currently it merely adds a test to reproduce faulty behavior reported by an NSD user.

### DIFF
--- a/ixfr.c
+++ b/ixfr.c
@@ -948,15 +948,12 @@ size_t ixfr_data_size(struct ixfr_data* data)
 }
 
 struct ixfr_store* ixfr_store_start(struct zone* zone,
-	struct ixfr_store* ixfr_store_mem, uint32_t old_serial,
-	uint32_t new_serial)
+	struct ixfr_store* ixfr_store_mem)
 {
 	struct ixfr_store* ixfr_store = ixfr_store_mem;
 	memset(ixfr_store, 0, sizeof(*ixfr_store));
 	ixfr_store->zone = zone;
 	ixfr_store->data = xalloc_zero(sizeof(*ixfr_store->data));
-	ixfr_store->data->oldserial = old_serial;
-	ixfr_store->data->newserial = new_serial;
 	return ixfr_store;
 }
 
@@ -1145,12 +1142,12 @@ static void store_soa(uint8_t* soa, struct zone* zone, uint32_t ttl,
 	write_uint32(sp, minimum);
 }
 
-void ixfr_store_add_newsoa(struct ixfr_store* ixfr_store,
-	struct buffer* packet, size_t ttlpos)
+void ixfr_store_add_newsoa(struct ixfr_store* ixfr_store, uint32_t ttl,
+	struct buffer* packet, size_t rrlen)
 {
 	size_t oldpos, sz = 0;
-	uint32_t ttl, serial, refresh, retry, expire, minimum;
-	uint16_t rdlen_uncompressed, rdlen_wire;
+	uint32_t serial, refresh, retry, expire, minimum;
+	uint16_t rdlen_uncompressed;
 	int primns_len = 0, email_len = 0;
 	uint8_t primns[MAXDOMAINLEN + 1], email[MAXDOMAINLEN + 1];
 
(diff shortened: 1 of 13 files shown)


k0ekk0ek:
Turns out the problem is not related zone verification, but is actually an issue with handling of IXFRs where consecutive updates are in the same IXFR and RRs added by the former are deleted again by the latter.

k0ekk0ek:
Last commit makes the test pass, but the code still needs a little more work. Took some time to simplify `apply_ixfr` too.

k0ekk0ek:
I've updated the ixfr_store code to set the `newserial` and `oldserial` members implicitly, which is more intuitive with the way the code is used in `apply_ixfr`. I've done the same for the code that creates IXFRs for consistency. I also moved checking for the first RR in `apply_ixfr` inside the loop. I'd rather have checking for first and second RRs in IXFR outside the loop, but the diff would change too much? Let me know if we want to merge it like this or you want other things changed.

ixfr.h
```diff
@@ -165,12 +164,10 @@ void ixfr_store_finish_data(struct ixfr_store* ixfr_store);
  * ixfr_store: stores ixfr data that is collected.
  * packet: DNS packet that contains the SOA. position restored on function
```
wcawijngaards, line 166: This documents the ttl parameter in the comment.
```suggestion
 * ttl: TTL of the SOA RR.
 * packet: DNS packet that contains the SOA. position restored on function
```

ixfr.c
```diff
@@ -1387,6 +1375,7 @@ int ixfr_store_add_newsoa_rdatas(struct ixfr_store* ixfr_store,
 	size_t capacity = 0;
 	if(ixfr_store->cancelled)
 		return 1;
+	ixfr_store->data->newserial = ntohl((uint32_t)*rdatas[2].data);
```
wcawijngaards, line 1386: Please check the rdata_num and the rdata_atom_size before using it like this, to prevent possible buffer length issues.
```suggestion
	if(rdata_num < 2 || rdata_atom_size(rdatas[2]) < 4)
		return 0;
	ixfr_store->data->newserial = ntohl((uint32_t)*rdatas[2].data);
```

ixfr.c
```diff
@@ -1443,6 +1432,22 @@ int ixfr_store_delrr_uncompressed(struct ixfr_store* ixfr_store,
...
+	for (size_t index=0; index < rdata_len; ) {
+		uint8_t label_size = rdata[index];
+		if (label_size == 0) {
+			return index + 1;
+		} else if ((label_size & 0xc0) != 0) {
+			index += 1;
+		} else {
+			index += label_size;
+		}
+	}
```
wcawijngaards: Perhaps it is useful to check that the rdata_len is sufficiently large that the information can be read from the buffer.

```suggestion
	for (size_t index=0; index < rdata_len; ) {
		uint8_t label_size = rdata[index];
		if (label_size == 0) {
			return index + 1;
		} else if ((label_size & 0xc0) != 0) {
			if(index + 1 >= rdata_len)
				return 0;
			index += 1;
		} else {
			if(index + label_size >= rdata_len)
				return 0;
			index += label_size;
		}
	}
```

wcawijngaards (approved):
This looks like good code for ixfr, with some rewrites in the apply_ixfr function. The test is good to have and looks straightforward.

ixfr.c
wcawijngaards: Perhaps this change is not optimal as the for condition already checks for it. It could be cautious, with more prominence on the boundary check, in the code, than it has currently. If you like it, perhaps.

ixfr.h
k0ekk0ek: Good catch. Updated.

ixfr.c
k0ekk0ek: I assumed the check would be implicit when the SOA rdata atoms are created. But yes, can't hurt to just check. Updated.

ixfr.c
k0ekk0ek: I'd like to keep this one as-is, but I did add a comment to clarify that if `index` exceeds `rdata_len`, `0` is automatically returned as the condition breaks the loop.

wcawijngaards: Perhaps there is another issue here, when there is a compression pointer it attempts to continue the loop. This is not really good, because it does not want to follow the dname compression pointer, it is attempting to skip the dname that is at this point, and so it should return after moving ahead one byte. So check if there is another byte, then skip that byte, and then return from the function, much like in the zero label case, since the dname has ended, in a compression pointer?

k0ekk0ek: Clearly I don't understand dname compression yet :sweat_smile: Well, we already know that there's another byte available as `index < rdata_len`, so that'd mean simply `return index + 1;` like with the zero label?

k0ekk0ek: or:
```
static size_t skip_dname(uint8_t* rdata, size_t rdata_len)
{   
  for (size_t index=0; index < rdata_len; ) {
    uint8_t label_size = rdata[index];
    if (label_size == 0 || (label_size & 0xc0) != 0) {
      return index + 1;
    } else {
      /* loop breaks if index exceeds rdata_len */
      index += label_size;
    }
  }
  
  return 0;
}
```

wcawijngaards: No, because it is one byte longer, check if that byte is in the rdata_len? The dname compression pointer is two bytes.

k0ekk0ek: so:
```c
static size_t skip_dname(uint8_t* rdata, size_t rdata_len)
{
  for (size_t index=0; index < rdata_len; ) {
    uint8_t label_size = rdata[index];
    if (label_size == 0) {
      return index + 1;
    } else if ((label_size & 0xc0) != 0) {
      return (index + 1 < rdata_len) ? index + 2 : 0;
    } else {
      /* loop breaks if index exceeds rdata_len */
      index += label_size;
    }
  }

  return 0;
}
```

k0ekk0ek: As the changes were already merged, I pushed the fix to master directly.

PR #301, opened 2023-10-31
improve the logging of ixfr fallbacks to axfr

ttyS4:
hi,

When nsd is used as a primary (like serving the zone after verification) it does not properly log if the zone transfer was ixfr or axfr.
It only logs what the client asked for. This is misleading because someone could think it was an ixfr reply, but with default configuration it falls back to axfr unless ixfr-out is explicitly configured.

This meant to fix that by properly logging when it falls back.

Regards,
 Tamás


wcawijngaards (approved):
These are good log output print statements. It should help pinpoint what IXFR and what AXFR answers are delivered, the verbosity level is similar to the level where notify packets are logged, so that is well chosen.

wcawijngaards:
Thank you for the code patch! The changes have been merged into the code.

PR #303, opened 2023-11-21
Add test to verify remaining AXFR are serviced

k0ekk0ek:


k0ekk0ek:
Test to verify is "incorrect". The timeout is set to 1/10th of a second when servicing remaining TCP connections. It's near impossible to ensure desired behavior is triggered from bash.

wcawijngaards (approved):
The change looks good and fixes bugs in tcp write of results.

PR #307, opened 2023-12-07
Many improvements to the nsd.conf man page

anandb-ripencc:
All references to the "database" option and database file name have been removed, save for one mention of the option, to say that it is ignored in NSD >= 4.8.0.

All references to "master" zones have been changed to "primary", and all references to "slave" zones have been changed to "secondary", in line with RFC 8499.

End-of-line spaces have been deleted from all lines. In some places, the text has been adjusted a little to read better.

### DIFF
--- a/nsd.conf.5.in
+++ b/nsd.conf.5.in
@@ -7,18 +7,17 @@
 .SH "SYNOPSIS"
 .B nsd.conf
 .SH "DESCRIPTION"
-.B Nsd.conf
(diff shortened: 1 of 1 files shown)


wcawijngaards (approved):
The textual changes look good! This is nice to have in the man page.

wcawijngaards:
Thanks for the text fixes, the zonesfiles-write option was incorrect due to the database option change and the fixes are good to have.

PR #309, opened 2023-12-07
More RFC 8499 compliance

anandb-ripencc:
Update nsd-control's output to use "secondary" instead of "slave".
Update the nsd-control man page to use "primary" and "secondary" instead of "master" and "slave".


wcawijngaards (approved):
The changes look good. These accurately update the terminology.

PR #315, opened 2024-01-29
XFR clients have to be able to query SOA at apex

wtoorop:
Clients that are allowed to transfer a zone because they match a provide-xfr acl, should also be able to query for the SOA at the apex (because such a client does that query before it starts the actual transfer).
This PR allows transfer clients to do a SOA query at the apex, even if queries are blocked for the zone with the `allow-query` option.

### DIFF
--- a/query.c
+++ b/query.c
@@ -1365,6 +1365,18 @@ answer_lookup_zone(struct nsd *nsd, struct query *q, answer_type *answer,
 				why->ip_address_spec,
 				why->nokey?"NOKEY":
 				(why->blocked?"BLOCKED":why->key_name)));
+		} else if(q->qtype == TYPE_SOA
+		       &&  0 == dname_compare(q->qname,
+				(const dname_type*)q->zone->opts->node.key)
+		       && -1 != acl_check_incoming(
+				q->zone->opts->pattern->provide_xfr, q,&why)) {
+			assert(why);
+			DEBUG(DEBUG_QUERY,1, (LOG_INFO, "SOA apex query %s "
+				"passed request-xfr acl %s %s",
+				dname_to_string(q->qname, NULL),
+				why->ip_address_spec,
+				why->nokey?"NOKEY":
+				(why->blocked?"BLOCKED":why->key_name)));
 		} else {
 			if (verbosity >= 2) {
 				char address[128];
--- a/tpkg/allow_query.tdir/allow_query.conf
+++ b/tpkg/allow_query.tdir/allow_query.conf
@@ -2,6 +2,9 @@
 server:
 	logfile: "nsd.log"
 	zonesdir: ""
+	username: ""
+	database: ""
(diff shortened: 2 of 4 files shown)


k0ekk0ek:
Why solve the problem this way? I mean, the logic is correct IMHO, but wouldn't it be easier to just add the address to the `allow_query` list when configuring? If the client queries for `AXFR` or `IXFR` it gets the entire zone anyway(?)

wtoorop:
> Why solve the problem this way? I mean, the logic is correct IMHO, but wouldn't it be easier to just add the address to the `allow_query` list when configuring? If the client queries for `AXFR` or `IXFR` it gets the entire zone anyway(?)

Part of it is convenience (or user friendliness). Not everybody may know a SOA query is needed for a transfer (or may have forgotten about it or not aware of it while configuring). It is also convenient that the TSIG requirements are copied from provide-xfr.
There is also a privacy enhancement when transfer is only allowed over TLS. An on path eavesdropper can otherwise still sniff the entire zone with spoofed UDP queries. Also, the TCP handshake (in the actual transfer) provides stronger guarantees that the sender is genuine.

k0ekk0ek:
I think there's some miscommunication? I meant: automatically add it to the `allow_query` acl when parsing the configuration file.

wtoorop:
> I think there's some miscommunication? I meant: automatically add it to the `allow_query` acl when parsing the configuration file.

Ah, I thought that you meant that the user could configure it herself.

My arguments w.r.t. privacy still holds though. We have an extra restriction to allow only TYPE_SOA at the apex queries with the provide-xfr acls, this restriction would not be there if the rules would simply be copied. The extra restriction prevents an on path eavesdropper to see the whole zone by sending queries wit spoofed source IP.

I also think the specific debugging message about the match mentioning provide-xfr has value.

k0ekk0ek:
My main concern is that it's executed in the hotpath and the acls are iterated in linear fashion(?) And if TSIG is not required the same party could simple send an AXFR query, which would be much easier than trying to traverse the entire zone with normal queries and guessing?

k0ekk0ek (approved):

wtoorop:
> My main concern is that it's executed in the hotpath and the acls are iterated in linear fashion(?) And if TSIG is not required the same party could simple send an AXFR query, which would be much easier than trying to traverse the entire zone with normal queries and guessing?

But that AXFR would be over TCP, yes? Then the TCP handshake prevents spoofing.

It's also not that much in the hotpath. The whole branch is only evaluated `if(q->zone->opts && q->zone->opts->pattern
&& q->zone->opts->pattern->allow_query)` on line 1330 and the provide-xfr acls only if the allow-query acls didn't allow the query in the first place.

wcawijngaards (approved):
The code looks good. The allowed SOA query should help with operations.

PR #319, opened 2024-03-27
Force install of auxiliary files in makedist.sh

k0ekk0ek:
What it says on the tin.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5,7 +5,7 @@ dnl
 sinclude(acx_nlnetlabs.m4)
 sinclude(dnstap/dnstap.m4)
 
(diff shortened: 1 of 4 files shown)


k0ekk0ek:
Turns out this doesn't work as expected. I was notified the shebang was changed for the install script. RC1 shipped with `#!/usr/bin/sh` while normally that is `#!/bin/sh`. I'm assuming it has to do with me testing this on a Linux workstation(?) Something to look into for the actual 4.9.0 release.

wcawijngaards:
For Unbound, say 1.19.3, the release ships with install-sh with `#!/usr/bin/sh` in it. That has no troubles, and also has not changed recently.

PR #322, opened 2024-04-04
Only copy missing auxiliary files

k0ekk0ek:
Remove the `-f` flag as that overrides the `install-sh` script. Now that the tarball is generated in a separate directory, only `-i` (copy missing standard auxiliary files) is needed, `-f` (consider all generated and standard files obsolete) is not.


k0ekk0ek:
Using `-f` results in the tarball containing a `install-sh~` file.

PR #328, opened 2024-04-22
Update README and pull latest simdzone

k0ekk0ek:
Update README.md to reflect changes in building from repository and use latest to simdzone to fix bug in name parser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ or post a message on the
 You can learn more about NSD by reading our
 [documentation](https://nsd.docs.nlnetlabs.nl/).
 
-## Compiling
+## Building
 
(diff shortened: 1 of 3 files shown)


wcawijngaards (approved):
This looks fine, there are changes needed for the simdzone package. Should configure be able to detect an installed libsimdzone?

k0ekk0ek:
We could, eventually. For now I'd like to stick with this since we may need to do some quick fixes in simdzone once people actively start using it in NSD. Let's wait for a little bit and then also support that use case?

wcawijngaards:
Yes, we can do that. Since the simdzone package is new at this time.

PR #331, opened 2024-05-15
Bugfix to deal with lost tasks after bad transfers

wtoorop:
The issue is that updating the database from transfers and other tasks that change the database (such as adding zones or changing patterns) are processed in so called task list. Especially applying transfers are executed batch-wise periodically as configured by `xfrd-reload-timeout:` (default every second). If a zone is added it is appended to the tasklist. If there would have been an "apply transfer task" before that other (adding a zone) task, then that will be executed first. If the "apply transfer task" fails, the reload will be aborted (because the database could be corrupt) and the effect of any tasks before or after the transfer are undone. This can lead to a situation that the transfer daemon (xfrd) thinks a zone is configured, but the zone is not configured in the server process (main).


tpkg/tasks_after_bad_xfr.tdir/tasks_after_bad_xfr.dsc
```diff
@@ -0,0 +1,18 @@
+BaseName: tasks_after_bad_xfr
+Version: 1.0
+Description: Catalog zones testing
```
wcawijngaards, line 3: Copy and paste mistake.
```suggestion
Description: Test tasks after bad transfers
```

wcawijngaards (approved):
The code looks nice. After reload step one, the `xfrd->reload_pid` variable is not assigned to a new value, where it would be set to the new reload process value in current code. After the reload completes the second step, the `xfrd->reload_pid` has a value, and that is fine. But just after step one it points to a deleted process, I think.

wtoorop:
Sorry @wcawijngaards and @k0ekk0ek if you already reviewed. I think this can be very much simplified. A new (better) proposal is in PR #332 .

PR #332, opened 2024-05-25
Make sure other tasks are processed with tasklists that contain both bad xfrs (that cause reload to exit) and other tasks

wtoorop:


wcawijngaards (approved):
The code looks okay. I worry that the tasks processed by the old-main process take a long time, and that old-main is not responsive during that time. For tasks for checkzonefiles or writezonefiles, that also seems reasonable.

wtoorop:
> The code looks okay. I worry that the tasks processed by the old-main process take a long time, and that old-main is not responsive during that time. For tasks for checkzonefiles or writezonefiles, that also seems reasonable.

Thanks @wcawijngaards commit [0c4e275](https://github.com/NLnetLabs/nsd/pull/332/commits/0c4e275b67f3139a5d56284f49cc30c6af8f3842) addresses this issue you raised of notifies not being passed when non transfer tasks are being processed (and it works), although it's hard to write a unit test for this (but I've tested manually with a strategically placed `sleep`). If you can review again and approve (perhaps only the last commit), then I can merge this as solution for the netnod case.

server.c
```diff
@@ -2863,7 +2917,99 @@ server_main(struct nsd *nsd)
...
+			}
+			udb_ptr_init(&xfrs2process, nsd->task[nsd->mytask]);
+			udb_ptr_init(&last_task   , nsd->task[nsd->mytask]);
+			reload_process_non_xfr_tasks(nsd, &xfrs2process
+							, &last_task);
+			if(n_non_xfr_tasks) {
+				sig_atomic_t cmd;
+
+				/* Send quit command to notifier, not waiting
+				 * for comfirmation.
```
wcawijngaards:
```suggestion
				 * for confirmation. 
```

wcawijngaards (approved):
Okay that looks nice and the server stays responsive to notifies. It creates another process, but it seems to also solve the responsiveness problem.

k0ekk0ek (approved):
I personally dislike the additional fork. IMHO it would be better to remove the `NOTIFY` proxy code from the main process and just let the serve child talk directly to xfrd, though there is probably a reason why it is the way it is too.

wtoorop:
> I personally dislike the additional fork. IMHO it would be better to remove the `NOTIFY` proxy code from the main process and just let the serve child talk directly to xfrd, though there is probably a reason why it is the way it is too.

Passing `NOTIFY` from serve childs to xfrd is developed in branch [devel/stream_notify2xfrd](https://github.com/NLnetLabs/nsd/tree/devel/stream_notify2xfrd).
PR #390 is this PR rolled back to commit [22addb5](https://github.com/NLnetLabs/nsd/pull/332/commits/22addb563feffa607c621ec012dd9487e93a09da) and with the [devel/stream_notify2xfrd](https://github.com/NLnetLabs/nsd/tree/devel/stream_notify2xfrd) merged in. It replaces this PR.

Issue #336, opened 2024-06-14
mutual tls authentication is not complete


bilias:
nsd.conf(5) specifies:
```
client-cert: <file name of clientcert.pem>
              If  you  want to use mutual TLS authentication, this is where the client certificates can be configured that NSD uses to connect to the upstream server to download the zone. The client public  key  pem  cert file can be configured here. Also configure a private key with client-key.

client-key: <file name of clientkey.key>
              If  you  want  to  use  mutual  TLS authentication, the private key file can be configured here for the client authentication.
```
So somehow mutual auth should work, but it doesn't

Client (slave/secondary) indeed verifies servers's (master/primary) certificate by using:
tls-cert-bundle: for the server's certificate (ca store)
tls-auth: auth-domain-name: for verifying the certificate CN.

However, server accepts whatever certificate is presented by the client (in the tls-auth sertting /client-cert/client-key)

Looking at the code:
SSL_CTX* server_tls_ctx_create(struct nsd *nsd, char* verifypem, char* ocspfile);
is called in nsd.c:1580
```
if(!(nsd.tls_ctx = server_tls_ctx_create(&nsd, NULL, 
      nsd.options->tls_service_ocsp)))
```
So, verifypem is always NULL, thus server does not perform any authentication for the client side of the connection. This is not mutual auth, only client verifies/authenticates the tls connection.

Either the manual page has to change, or maybe put an extra tls-auth field in
provide-xfr: to force tls-auth for the client
This can by done again by looking at 
tls-cert-bundle: for the client's certificate
tls-auth: auth-domain-name: for verifying the certificate CN.

Ideally each tls-auth could also include server ca-pem/cert-bundle dinstict from global tls-cert-bundle. You don't want clientA to be able to authenticate with clientB's certificate/key on server even if both are valid and listed in tls-cert-bundle. 

Can provide test setup on localhost that shows the issue clearly.

k0ekk0ek:
Thanks again @bilias! With the changes merged, I think this can be closed?

bilias:
@k0ekk0ek 
Yes, both sides of TLS-AUTH were covered by the merge. Can be closed.

PR #337, opened 2024-06-15
Mutual TLS-AUTH

bilias:
First part: Client side TLS-auth (client is authenticating to server).
[client's authentication of the server is fine]

commit 10eca881c6b27d62d95d0eee97f67bb08a0a632c

When request-xfr is setup on secondary with tls-auth and client-cert/client-key defined,
client/secondary makes initial connection to server/primary without having setup it's client-cert and client-key at an early stage.
The order that is done on the client is:
SSL_CTX_new
SSL_CTX_set_default_verify_paths
SSL_CTX_load_verify_locations

SSL_new()
SSL_set_connect_state()
SSL_set_mode(tp->ssl, SSL_MODE_AUTO_RETRY);
SSL_set_verify(tp->ssl, SSL_VERIFY_PEER, tls_verify_callback);
SSL_set1_host(tp->ssl, auth_domain_name)
SSL_CTX_use_certificate_chain_file()
SSL_CTX_use_PrivateKey_file()
ssl_handshake()

So the first SSL_new() does not have certificate_chain_file/PrivateKey_file which are loaded on CTX and not inherited in that first SSL_new().

Subsequent connections have the key but the first connection does not.

If server is setup to require/force client cert/key, on server we see:
error: TLS handshake failed from 127.0.0.1 crypto error:0A0000C7:SSL routines::peer did not return a certificate

If we load first the cert/key on the client and then make connection to server,
client will reply to server's request for client authentication. I've just changed the order things are done.

This is the first patch, for the client side, in order for the client to authenticate correctly to server using tls-auth.
I'm preparing another patch for the server.c in order for the server to require/force client to authenticate on a provide-xfr using tls-auth and complete mutual auth. 
#336

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1045,7 +1045,7 @@ if test x$HAVE_SSL = x"yes"; then
 	fi
 	SSL_LIBS="-lssl"
 	AC_SUBST(SSL_LIBS)
-	AC_CHECK_HEADERS([openssl/ssl.h openssl/err.h openssl/rand.h openssl/ocsp.h openssl/core_names.h],,, [AC_INCLUDES_DEFAULT])
+	AC_CHECK_HEADERS([openssl/ssl.h openssl/err.h openssl/rand.h openssl/ocsp.h openssl/core_names.h openssl/x509v3.h],,, [AC_INCLUDES_DEFAULT])
 	AC_CHECK_FUNCS([HMAC_CTX_reset HMAC_CTX_new EVP_cleanup ERR_load_crypto_strings OPENSSL_init_crypto CRYPTO_memcmp EC_KEY_new_by_curve_name EVP_MAC_CTX_new EVP_MAC_CTX_set_params EVP_MAC_CTX_get_mac_size SHA1_Init])
 	if test "$ac_cv_func_SHA1_Init" = "yes"; then
 		ACX_FUNC_DEPRECATED([SHA1_Init], [(void)SHA1_Init(NULL);], [
@@ -1071,6 +1071,9 @@ AC_INCLUDES_DEFAULT
 #endif
 #include <openssl/ssl.h>
 #include <openssl/evp.h>
+#ifdef HAVE_OPENSSL_X509V3_h
+#include <openssl/x509v3.h>
+#endif
 ])
 	AC_CHECK_DECL([TLS1_3_VERSION], 
 		[AC_DEFINE([HAVE_TLS_1_3], [1], [Define if TLS 1.3 is supported by OpenSSL])],
@@ -1078,7 +1081,7 @@ AC_INCLUDES_DEFAULT
 
 	BAKLIBS="$LIBS"
 	LIBS="-lssl $LIBS"
-	AC_CHECK_FUNCS([OPENSSL_init_ssl SSL_get1_peer_certificate SSL_CTX_set_security_level ERR_load_SSL_strings])
+	AC_CHECK_FUNCS([OPENSSL_init_ssl SSL_get1_peer_certificate SSL_CTX_set_security_level X509_get_ext_d2i OPENSSL_sk_num OPENSSL_sk_value OPENSSL_sk_pop_free ASN1_STRING_get0_data ASN1_STRING_length X509_get_subject_name X509_NAME_get_index_by_NID X509_NAME_get_entry X509_NAME_ENTRY_get_data ERR_load_SSL_strings])
 	if test "$ac_cv_func_ERR_load_SSL_strings" = "yes"; then
 		ACX_FUNC_DEPRECATED([ERR_load_SSL_strings], [(void)ERR_load_SSL_strings();], [
 #include <openssl/ssl.h>
--- a/configure.ac
+++ b/configure.ac
@@ -1081,7 +1081,7 @@ AC_INCLUDES_DEFAULT
 
 	BAKLIBS="$LIBS"
 	LIBS="-lssl $LIBS"
-	AC_CHECK_FUNCS([OPENSSL_init_ssl SSL_get1_peer_certificate SSL_CTX_set_security_level X509_get_ext_d2i OPENSSL_sk_num OPENSSL_sk_value OPENSSL_sk_pop_free ASN1_STRING_get0_data ASN1_STRING_length X509_get_subject_name X509_NAME_get_index_by_NID X509_NAME_get_entry X509_NAME_ENTRY_get_data ERR_load_SSL_strings])
+	AC_CHECK_FUNCS([OPENSSL_init_ssl SSL_get1_peer_certificate SSL_CTX_set_security_level ERR_load_SSL_strings])
 	if test "$ac_cv_func_ERR_load_SSL_strings" = "yes"; then
 		ACX_FUNC_DEPRECATED([ERR_load_SSL_strings], [(void)ERR_load_SSL_strings();], [
 #include <openssl/ssl.h>
--- a/options.c
+++ b/options.c
@@ -14,6 +14,10 @@
 #ifdef HAVE_IFADDRS_H
 #include <ifaddrs.h>
 #endif
+#ifdef HAVE_SSL
+#include <openssl/ssl.h>
+#include <openssl/x509v3.h>
+#endif
 #include "options.h"
 #include "query.h"
 #include "tsig.h"
@@ -290,6 +294,13 @@ parse_options_file(struct nsd_options* opt, const char* file,
 		}
 		for(acl=pat->provide_xfr; acl; acl=acl->next)
 		{
+			/* Find tls_auth */
+			if (!acl->tls_auth_name)
+				; /* pass */
+			else if (!(acl->tls_auth_options =
+			                tls_auth_options_find(opt, acl->tls_auth_name)))
+				c_error("tls_auth %s in pattern %s could not be found",
+						acl->tls_auth_name, pat->pname);
 			if(acl->nokey || acl->blocked)
 				continue;
 			acl->key_options = key_options_find(opt, acl->key_name);
@@ -1940,6 +1951,12 @@ acl_check_incoming(struct acl_options* acl, struct query* q,
 
 	while(acl)
 	{
+#ifdef HAVE_SSL
+		DEBUG(DEBUG_XFRD,2, (LOG_INFO, "testing acl %s %s %s",
+			acl->ip_address_spec, acl->nokey?"NOKEY":
+			(acl->blocked?"BLOCKED":acl->key_name),
+			(acl->tls_auth_name && q->tls_auth)?acl->tls_auth_name:""));
+#endif
 		DEBUG(DEBUG_XFRD,2, (LOG_INFO, "testing acl %s %s",
 			acl->ip_address_spec, acl->nokey?"NOKEY":
 			(acl->blocked?"BLOCKED":acl->key_name)));
@@ -1955,6 +1972,28 @@ acl_check_incoming(struct acl_options* acl, struct query* q,
 				return -1;
 			}
 		}
+#ifdef HAVE_SSL
+		/* we are in a acl with tls_auth */
+		if (acl->tls_auth_name && q->tls_auth) {
+			/* we have auth_domain_name in tls_auth */
+			if (acl->tls_auth_options && acl->tls_auth_options->auth_domain_name) {
+				q->cert_cn = NULL;
+				if (!acl_tls_hostname_matches(q->tls_auth, acl->tls_auth_options->auth_domain_name, &(q->cert_cn))) {
+					VERBOSITY(3, (LOG_WARNING,
+							"client cert %s does not match %s %s",
+							q->cert_cn, acl->tls_auth_name, acl->tls_auth_options->auth_domain_name));
+					q->cert_cn = NULL;
+					return -1;
+				}
+				VERBOSITY(5, (LOG_INFO, "%s %s verified",
+					acl->tls_auth_name, acl->tls_auth_options->auth_domain_name));
+				q->cert_cn = acl->tls_auth_options->auth_domain_name;
+			} else {
+				/* nsd gives error on start for this, but check just in case */
+				log_msg(LOG_ERR, "auth-domain-name not defined in %s", acl->tls_auth_name);
+			}
+		}
+#endif
 		number++;
 		acl = acl->next;
 	}
@@ -2157,6 +2196,146 @@ acl_addr_match_range_v6(uint32_t* minval, uint32_t* x, uint32_t* maxval, size_t 
 	return 1;
 }
 #endif /* INET6 */
+
+#ifdef HAVE_SSL
+int
+acl_tls_hostname_matches(SSL* tls_auth, const char* acl_cert_cn, char** cert_cn)
+{
+	X509 *client_cert;
+	*cert_cn = NULL;
+
+#  ifdef HAVE_SSL_GET1_PEER_CERTIFICATE
+	if ((client_cert = SSL_get1_peer_certificate(tls_auth)) == NULL) {
+#  else
+	if ((client_cert = SSL_get_peer_certificate(tls_auth)) == NULL) {
+#  endif
+		DEBUG(DEBUG_XFRD,2, (LOG_INFO, "CN match fail no peer certificate"));
+		X509_free(client_cert);
+		return 0;
+	}
+	if (!acl_cert_cn || !client_cert) {
+		X509_free(client_cert);
+		return 0;
+	}
+
+	/* OpenSSL provides functions for hostname checking from certificate
+	 * Following code should work but it doesn't.
+	 * Keep it for future test in order to not use custom code
+	 *
+	 * X509_VERIFY_PARAM *vpm = SSL_get0_param(tls_auth);
+	 * Hostname check is done here:
+	 * X509_VERIFY_PARAM_set1_host(vpm, acl_cert_cn, 0); // recommended
+	 * X509_check_host() // can also be used instead. Not recommended DANE-EE
+	 * SSL_get_verify_result(tls_auth) != X509_V_OK) // NOT ok
+	 * const char *peername = X509_VERIFY_PARAM_get0_peername(vpm); // NOT ok
+	 */
+
+/* Code in rest of function is from
+ * https://wiki.openssl.org/index.php/Hostname_validation
+ * with modifications
+ *
+ * Obtained from: https://github.com/iSECPartners/ssl-conservatory
+ * Copyright (C) 2012, iSEC Partners.
+ * License: MIT License
+ * Author:  Alban Diquet
+ */
+
+	/* semi follow RFC6125#section-6.4.4 check SAN DNS first */
+	int i;
+	int san_names_nb = -1;
+	STACK_OF(GENERAL_NAME) *san_names = NULL;
+
+	/* Try to extract the names within the SAN extension from the certificate */
+	san_names = X509_get_ext_d2i(client_cert, NID_subject_alt_name, NULL, NULL);
+	if (san_names != NULL) {
+		san_names_nb = sk_GENERAL_NAME_num(san_names);
+
+		/* Check each name within the extension */
+		for (int i = 0; i < san_names_nb; i++) {
+			const GENERAL_NAME *current_name = sk_GENERAL_NAME_value(san_names, i);
+			/* skip non-DNS SAN entries */
+			if (current_name->type != GEN_DNS)
+				continue;
+			char *dns_name = (char *) ASN1_STRING_get0_data(current_name->d.dNSName);
+			if (dns_name == NULL)
+				continue;
+			/* Make sure there isn't an embedded NUL character in the DNS name */
+			if ((size_t)ASN1_STRING_length(current_name->d.dNSName) != strlen(dns_name)) {
+				DEBUG(DEBUG_XFRD,2, (LOG_WARNING, "SAN Malformed certificate"));
+				break;
+			}
+			if (dns_name != NULL) {
+				/* certificate SAN DNS match */
+				if (strncmp(dns_name, acl_cert_cn, strlen(acl_cert_cn))==0) {
+					DEBUG(DEBUG_XFRD,2, (LOG_INFO, "SAN %s matches acl", dns_name));
+					*cert_cn = strndup(dns_name, strlen(dns_name));
+					X509_free(client_cert);
+					return 1;
+				} else {
+					DEBUG(DEBUG_XFRD,2, (LOG_INFO, "SAN %s does not match acl", dns_name));
+					/* check with remaining SANs and then continue with normal CN check */
+				}
+			}
+		}
+		sk_GENERAL_NAME_pop_free(san_names, GENERAL_NAME_free);
+	} else {
+		DEBUG(DEBUG_XFRD,2, (LOG_INFO, "No SAN DNS extension in certificate"));
+	}
+
+	/* no match on SAN, continue with normal CN */
+	int common_name_loc = -1;
+	char *common_name_str = NULL;
+	X509_NAME *subject_name = NULL;
+	X509_NAME_ENTRY *common_name_entry = NULL;
+	ASN1_STRING *common_name_asn1 = NULL;
+
+	if ((subject_name = X509_get_subject_name(client_cert)) == NULL) {
+		DEBUG(DEBUG_XFRD,2, (LOG_INFO, "CN get subject failed"));
+		X509_free(client_cert);
+		return 0;
+	}
+	/* Find the position of the CN field in the Subject field of the certificate */
+	common_name_loc = X509_NAME_get_index_by_NID(subject_name, NID_commonName, -1);
+	if (common_name_loc < 0) {
+		DEBUG(DEBUG_XFRD,2, (LOG_INFO, "CN not found in Subject"));
+		X509_free(client_cert);
+		return 0;
+	}
+	/* Extract the CN field */
+	common_name_entry = X509_NAME_get_entry(subject_name, common_name_loc);
+	if (common_name_entry == NULL) {
+		DEBUG(DEBUG_XFRD,2, (LOG_INFO, "CN extraction failed"));
+		X509_free(client_cert);
+		return 0;
+	}
+	/* Convert the CN field to a C string */
+	common_name_asn1 = X509_NAME_ENTRY_get_data(common_name_entry);
+	if (common_name_asn1 == NULL) {
+		DEBUG(DEBUG_XFRD,2, (LOG_INFO, "CN name to string convertion failed"));
+		X509_free(client_cert);
+		return 0;
+	}
+	common_name_str = (char *) ASN1_STRING_get0_data(common_name_asn1);
+	/* Make sure there isn't an embedded NUL character in the CN */
+	if ((size_t)ASN1_STRING_length(common_name_asn1) != strlen(common_name_str)) {
+		DEBUG(DEBUG_XFRD,2, (LOG_WARNING, "CN Malformed certificate"));
+		X509_free(client_cert);
+		return 0;
+	}
+	if (common_name_str != NULL) {
+		/* store it for logging */
+		*cert_cn = strndup(common_name_str, strlen(common_name_str));
+		/* certificate CN match */
+		if (strncmp(common_name_str, acl_cert_cn, strlen(acl_cert_cn))==0) {
+			X509_free(client_cert);
+			return 1;
+		}
+	}
+	DEBUG(DEBUG_XFRD,2, (LOG_INFO, "CN from cert %s does not match acl", common_name_str));
+	X509_free(client_cert);
+	return 0;
+}
+#endif
 
 int
 acl_key_matches(struct acl_options* acl, struct query* q)
--- a/options.c
+++ b/options.c
@@ -18,6 +18,9 @@
 #include <openssl/ssl.h>
 #include <openssl/x509v3.h>
 #endif
+#ifdef HAVE_STRINGS_H
+#include <strings.h>
+#endif
 #include "options.h"
 #include "query.h"
 #include "tsig.h"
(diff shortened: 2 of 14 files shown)


bilias:
Second part (server authenticates the client):

The second commit, adds a dedicated port on the server in order to enforce TLS Authentication for connecting clients (secondary servers) and address #336 from the server's side.

nsd does not support multiple tls-port(s) so I had to put checks in multiple places to see if we are on tls-port or tls-auth-port.
Not the ideal way of doing this but it works.

Another approach would be to put a tls-auth: yes/no knob where we could enforce tls authentication on the normal tls-port.
This would be a much more simple solution. Happy to make it if this one is not accepted.

Now let me explain a little about my initial approach which was adding tls-auth on provide-xfr acl.
TLS handshake and the creation on nsd.tls_ctx and data->tls are created an a stage where we don't have access yet to the provide-xfr ACL in order to check if we require tls-auth.

I've tried the approach of doing SSL_CTX_set_verify with SSL_VERIFY_PEER | SSL_VERIFY_POST_HANDSHAKE
on the server in order to make the initial TLS handshake and request for a client authenticate at a later stage.

Client sends SSL_set_post_handshake_auth() in the first ClientHello to say that it supports Post-Handshake Authentication (PHA)

Server carries data->tls pointer down to query_process() where it calls a new function process_tls_auth() in order to 
call SSL_verify_client_post_handshake() and issue another TLS handshake saying that it now requires client authentication.

query_process() was the first place where the ACL requiring tls-auth was available to me.
Note that PHA is only supposed to work for TLS1.3. TLS1.2 does renegotiation which is not available in TLS1.3

```
static nsd_rc_type
process_tls_auth(SSL *tls, struct query* q, nsd_type *nsd)
{
   struct zone_options* zone_opt;
   zone_opt = zone_options_find(nsd->options, q->qname);

   if (tls && zone_opt->pattern->provide_xfr->tls_auth_name) {
      // code taken from s_server.c from openssl
      SSL_set_verify(tls, SSL_VERIFY_PEER, NULL);
      if (!SSL_verify_client_post_handshake(tls)) {
         log_msg(LOG_ERR, "error in SSL_verify_client_post_handshake");
         return NSD_RC_REFUSE;
      } else {
         if(!SSL_do_handshake(tls)) {
            log_msg(LOG_ERR, "FAILED: SSL_do_handshake");
            return NSD_RC_REFUSE; // maybe not this, but fail in any case
         } else {
            log_msg(LOG_INFO, "TLS-AUTH SUCCESS!");
         }
      }
   }
   return NSD_RC_OK;
}
```
Up to this point I knew if we are on an ACL that requires TLS-AUTH in order to provide XFR to that specific client.


Don't know if the problem was calling SSL_do_handshake direct and not via tls_handshake() but it didn't work.
event code and handlers were complicated for me to play more with this.

This is where I stopped and though about tls-auth-port as a dedicated SSL_CTX that requires
SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT

As I said, since this is the enforcing option, as well as passing a non-empty CAfile, this could also be done via a simpler YES/NO knob to enforce it on normal tls-port.

wcawijngaards:
The `NSD_RC_NOTAUTH` return is not right, it means not authoritative for the zone, and to communicate that access is not granted, I think `NSD_RC_REFUSE` is a better choice of DNS RCODE value.

If the client has to set `SSL_set_post_handshake_auth` before connection starts, then perhaps it is nicer to support it somehow where the client action is not needed. With regular client authentication, this realistically more for compatibility, because I do not think any software sets that before the connection.

It is nicer if this option can exist together with regular tls service, so an on/off button is not really where I would take it either. But the check of client certificate at the initial handshake stage, on the tls_auth port, seems like a fine idea to me. If that works for the clients.

bilias:
> If the client has to set `SSL_set_post_handshake_auth` before connection starts, then perhaps it is nicer to support it somehow where the client action is not needed. With regular client authentication, this realistically more for compatibility, because I do not think any software sets that before the connection.

I agree that most clients do not set it. There are multiple talks online about this and various browsers that do not want to set it. Don't know about resolvers.

> It is nicer if this option can exist together with regular tls service, so an on/off button is not really where I would take it either. But the check of client certificate at the initial handshake stage, on the tls_auth port, seems like a fine idea to me. If that works for the clients.

I'm not 100% decided that an on/off knob is that bad.
nsd is an authoritative server. Why would you need TLS in the first place except for zone transfers?
I can't think of a scenario where you want some clients to authenticate (on the TLS port) and some others to not.

Maybe if you want different policies per zone (stubs for instance). But then you need to go down to the ACL and tls-auth.

In any case the tls-auth-port seems the most friendly one in terms of compatibility and it does not sacrifice/alter the normal TLS behaviour.

bilias:
As a side note, if this is finally accepted, I would also add something like the following in the manual page, which is important in my opinion:

Client's certificate name in not validated by the server. Server accepts any certificate presented by the client that is validated with tls-cert-bundle. If server uses the default verify location of the system, then any certificate signed by a valid CA is accepted. For this reason it is highly recommended to either setup a local PKI or create self signed certificates specifically for this purpose (XoT) for your secondary servers and only add those to a custom tls-cert-bundle file on the primary.
In advance primary should also filter (firewall) connections from secondaries to other listening ports in order to prohibit unauthenticated zone transfers and specifically only allow them via the tls-auth-port.

bilias:
@wcawijngaards do you know if this will be accepted?

If yes, I would like to extend my work on top of this and address the 2 issues I've mentioned above.

>In advance primary should also filter (firewall) connections from secondaries to other listening ports in order to prohibit unauthenticated zone transfers and specifically only allow them via the tls-auth-port.

This can be solved on the nsd side, by implementing something like a "tls-auth-xfr-only" yes/no knob
If yes, only xfrs that are in data->tls_auth (authenticated) will be accepted

> Client's certificate name in not validated by the server. Server accepts any certificate presented by the client that is validated with tls-cert-bundle. If server uses the default verify location of the system, then any certificate signed by a valid CA is accepted. For this reason it is highly recommended to either setup a local PKI or create self signed certificates specifically for this purpose (XoT) for your secondary servers and only add those to a custom tls-cert-bundle file on the primary. 

This can be solved by adding "tls-auth" in provide-xfr like it is in request-xfr, and use auth-domain-name from specified tls-auth, in order to verify client name is matching the presented certificate. The same way client verifies server's certificate.

Interested on these? I believe they are not hard to implement and will add to the security.

Now, for both of these to work
data->tls and
data->tls_auth must go from handle_tcp_reading()/handle_tls_reading() down to query_process()

I can pass both of them as args or the whole data struct pointer. What do you think is best?

wcawijngaards:
I would like to have this accepted. For passing the parameter, it is now part of tcp_data struct, but at line server.c:5293, over there, it can be stored as data->query->variable = variable_to_save; and then in the query handler it can use query->variable to access the information. The query->variable is then added to struct query in query.h. Perhaps that can help have the information accessible that is needed.

bilias:
> I would like to have this accepted. For passing the parameter, it is now part of tcp_data struct, but at line server.c:5293, over there, it can be stored as data->query->variable = variable_to_save; and then in the query handler it can use query->variable to access the information. The query->variable is then added to struct query in query.h. Perhaps that can help have the information accessible that is needed.

Thanks,
I believe it's better to go in two steps. So take this one in when you feel ok with it and I'll prepare a new push request since it is actually a new feature (on top of this one). It will be easier if something comes up.

If I understand correctly, you're saying I should extend struct query and add the info needed there?
Will probably need just those 2 from struct tcp_handler_data
SSL* tls;
SSL* tls_auth;

bilias:
@wcawijngaards 
I've added tls-auth-xfr-only to allow XFR only on the tls-auth-port and to authenticated clients only

If you want to have a peak:
https://github.com/bilias/nsd/tree/tls-auth-xfr-only
compare it to my master tree to see this change only.

Next is client name verification.

bilias:
Everything is completed from my side.
Final repo with all features is https://github.com/bilias/nsd/tree/tls_auth

Do you want me to do a new PR with everything inside or do it in parts?
Each feature is on a different branch so we split them.

bilias:
Everything is ready and uploaded into this PR but in different commits.

1) a87e820577ed1975be42f0a2cae378c86a97a06a
Fixes a bug in client side where the first connection to server was without SSL. First load cert/key then connect

2) https://github.com/NLnetLabs/nsd/pull/337/commits/edeca999a04259c7d420baa5397487f1cb604ac7
Adds a dedicated tls-auth-port where connections/XFRS must come from authenticated clients

3) https://github.com/NLnetLabs/nsd/pull/337/commits/187f0930676301a223c4dfc8a68c50e3ae750d68
Adds tls-auth-xfr-only option to allow XFRS only on tls-auth-port and only from authenticated clients. Works like "tls firewall"

4)  https://github.com/NLnetLabs/nsd/pull/337/commits/1fb6ed02a190ce1c8759e7dbeff35694327333ff
Adds tls-auth in provide-xfr ACL. Server validated the hostname of the connectinig client according to tls-auth settings.
This is to restrict valid certificates

bilias:
Everything is ready and uploaded into this PR but in different commits.

1) a87e820577ed1975be42f0a2cae378c86a97a06a
Fixes a bug in client side where the first connection to server was without SSL. First load cert/key then connect

2) edeca999a04259c7d420baa5397487f1cb604ac7
Adds a dedicated tls-auth-port where connections/XFRS must come from authenticated clients

3) 187f0930676301a223c4dfc8a68c50e3ae750d68
Adds tls-auth-xfr-only option to allow XFRS only on tls-auth-port and only from authenticated clients. Works like "tls firewall"

4)  1fb6ed02a190ce1c8759e7dbeff35694327333ff
Adds tls-auth in provide-xfr ACL. Server validated the hostname of the connectinig client according to tls-auth settings.
This is to restrict valid certificates

Future work: These kind of protections can very easily be extended to other connections as well, such as notify, in order for all commucations between primaries/secondaries to be protected by TLS with authentication based on certificates.

bilias:
Everything is ready and uploaded into this PR but in different commits.

1) a87e820577ed1975be42f0a2cae378c86a97a06a
Fixes a bug in client side where the first connection to server was without SSL. First load cert/key then connect

2) edeca999a04259c7d420baa5397487f1cb604ac7
Adds a dedicated tls-auth-port where connections/XFRS must come from authenticated clients

3) 187f0930676301a223c4dfc8a68c50e3ae750d68
Adds tls-auth-xfr-only option to allow XFRS only on tls-auth-port and only from authenticated clients. Works like "tls firewall"

4)  1fb6ed02a190ce1c8759e7dbeff35694327333ff
Adds tls-auth in provide-xfr ACL. Server validated the hostname of the connectinig client according to tls-auth settings.
This is to restrict valid certificates

Future work: These kind of protections can very easily be extended to other connections as well, such as notify, in order for all commucations between primaries/secondaries to be protected by TLS with authentication based on certificates.

bilias:
Everything is ready and uploaded into this PR but in different commits.

1) a87e820577ed1975be42f0a2cae378c86a97a06a
Fixes a bug in client side where the first connection to server was without SSL. First load cert/key then connect

2) edeca999a04259c7d420baa5397487f1cb604ac7
Adds a dedicated tls-auth-port where connections/XFRS must come from authenticated clients

3) 187f0930676301a223c4dfc8a68c50e3ae750d68
Adds tls-auth-xfr-only option to allow XFRS only on tls-auth-port and only from authenticated clients. Works like "tls firewall"

4)  a91e67c553b2933072d3238a098e8ef65379b143
Adds tls-auth in provide-xfr ACL. Server validates the hostname of the connecting client according to tls-auth settings.
This is to restrict valid certificates and not to accept any cert signed by a valid CA

Future work: These kind of protections can very easily be extended to other connections as well, such as notify, in order for all commucations between primaries/secondaries to be protected by TLS with authentication based on certificates.
This could also work for clients wanting normal traffic, irrelevant to zone transfers.

configure.ac
```diff
@@ -1071,14 +1071,17 @@ AC_INCLUDES_DEFAULT
...
+#endif
 ])
 	AC_CHECK_DECL([TLS1_3_VERSION], 
 		[AC_DEFINE([HAVE_TLS_1_3], [1], [Define if TLS 1.3 is supported by OpenSSL])],
 		[AC_MSG_WARN([No TLS 1.3, therefore XFR-over-TLS is disabled])], [[#include <openssl/ssl.h>]])
 
 	BAKLIBS="$LIBS"
 	LIBS="-lssl $LIBS"
-	AC_CHECK_FUNCS([OPENSSL_init_ssl SSL_get1_peer_certificate SSL_CTX_set_security_level ERR_load_SSL_strings])
+	AC_CHECK_FUNCS([OPENSSL_init_ssl SSL_get1_peer_certificate SSL_CTX_set_security_level X509_get_ext_d2i OPENSSL_sk_num OPENSSL_sk_value OPENSSL_sk_pop_free ASN1_STRING_get0_data ASN1_STRING_length X509_get_subject_name X509_NAME_get_index_by_NID X509_NAME_get_entry X509_NAME_ENTRY_get_data ERR_load_SSL_strings])
```
wcawijngaards: These checks are for functions that are now used, but since there are no ifdefs on the output of the checks, they are unnecessary to perform here.

```suggestion
	AC_CHECK_FUNCS([OPENSSL_init_ssl SSL_get1_peer_certificate SSL_CTX_set_security_level ERR_load_SSL_strings])
```

options.c
```diff
@@ -14,6 +14,10 @@
 #ifdef HAVE_IFADDRS_H
 #include <ifaddrs.h>
 #endif
+#ifdef HAVE_SSL
+#include <openssl/ssl.h>
+#include <openssl/x509v3.h>
+#endif
```
wcawijngaards, line 20: If the suggested strcasecmp routine is used, from another review comment, then the header include is useful to have.

```suggestion
#endif
#ifdef HAVE_STRINGS_H
#include <strings.h>
#endif
```

options.c
```diff
@@ -290,6 +294,13 @@ parse_options_file(struct nsd_options* opt, const char* file,
 		}
 		for(acl=pat->provide_xfr; acl; acl=acl->next)
 		{
+			/* Find tls_auth */
+			if (!acl->tls_auth_name)
+				; /* pass */
+			else if (!(acl->tls_auth_options =
+			                tls_auth_options_find(opt, acl->tls_auth_name)))
+				c_error("tls_auth %s in pattern %s could not be found",
+						acl->tls_auth_name, pat->pname);
```
wcawijngaards: Without the awkward no action then-clause, it is prettier.

```suggestion
			if (acl->tls_auth_name) {
				if (!(acl->tls_auth_options =
			                tls_auth_options_find(opt, acl->tls_auth_name)))
				    c_error("tls_auth %s in pattern %s could not be found",
						acl->tls_auth_name, pat->pname);
			}
```

options.c
```diff
@@ -1940,6 +1951,12 @@ acl_check_incoming(struct acl_options* acl, struct query* q,
...
 	{
+#ifdef HAVE_SSL
+		DEBUG(DEBUG_XFRD,2, (LOG_INFO, "testing acl %s %s %s",
+			acl->ip_address_spec, acl->nokey?"NOKEY":
+			(acl->blocked?"BLOCKED":acl->key_name),
+			(acl->tls_auth_name && q->tls_auth)?acl->tls_auth_name:""));
+#endif
 		DEBUG(DEBUG_XFRD,2, (LOG_INFO, "testing acl %s %s",
 			acl->ip_address_spec, acl->nokey?"NOKEY":
 			(acl->blocked?"BLOCKED":acl->key_name)));
```
wcawijngaards: With an else clause to not print the information twice, like in the other debug output prints.

```suggestion
#ifdef HAVE_SSL
		DEBUG(DEBUG_XFRD,2, (LOG_INFO, "testing acl %s %s %s",
			acl->ip_address_spec, acl->nokey?"NOKEY":
			(acl->blocked?"BLOCKED":acl->key_name),
			(acl->tls_auth_name && q->tls_auth)?acl->tls_auth_name:""));
#else
		DEBUG(DEBUG_XFRD,2, (LOG_INFO, "testing acl %s %s",
			acl->ip_address_spec, acl->nokey?"NOKEY":
			(acl->blocked?"BLOCKED":acl->key_name)));
#endif
```

options.c
```diff
@@ -1955,6 +1972,28 @@ acl_check_incoming(struct acl_options* acl, struct query* q,
...
+		/* we are in a acl with tls_auth */
+		if (acl->tls_auth_name && q->tls_auth) {
+			/* we have auth_domain_name in tls_auth */
+			if (acl->tls_auth_options && acl->tls_auth_options->auth_domain_name) {
+				q->cert_cn = NULL;
+				if (!acl_tls_hostname_matches(q->tls_auth, acl->tls_auth_options->auth_domain_name, &(q->cert_cn))) {
+					VERBOSITY(3, (LOG_WARNING,
+							"client cert %s does not match %s %s",
+							q->cert_cn, acl->tls_auth_name, acl->tls_auth_options->auth_domain_name));
+					q->cert_cn = NULL;
```
wcawijngaards, line 1990: The cert_cn is a memory leak, if it was returned with strdup from the acl_tls_hostname_matches routine. The string needs to be freed.

```suggestion
					free(q->cert_cn);
					q->cert_cn = NULL;
```

options.c
```diff
@@ -1955,6 +1972,28 @@ acl_check_incoming(struct acl_options* acl, struct query* q,
...
+					return -1;
+				}
+				VERBOSITY(5, (LOG_INFO, "%s %s verified",
+					acl->tls_auth_name, acl->tls_auth_options->auth_domain_name));
+				q->cert_cn = acl->tls_auth_options->auth_domain_name;
```
wcawijngaards, line 1996: The cert_cn is a memory leak, if it was returned with strdup from the acl_tls_hostname_matches routine. The string needs to be freed.

```suggestion
				free(q->cert_cn);
				q->cert_cn = acl->tls_auth_options->auth_domain_name;
```

options.c
```diff
@@ -2158,6 +2197,146 @@ acl_addr_match_range_v6(uint32_t* minval, uint32_t* x, uint32_t* maxval, size_t
...
+	if (common_name_str != NULL) {
+		/* store it for logging */
+		*cert_cn = strndup(common_name_str, strlen(common_name_str));
+		/* certificate CN match */
+		if (strncmp(common_name_str, acl_cert_cn, strlen(acl_cert_cn))==0) {
+			X509_free(client_cert);
+			return 1;
+		}
+	}
+	DEBUG(DEBUG_XFRD,2, (LOG_INFO, "CN from cert %s does not match acl", common_name_str));
```
wcawijngaards: The `common_name_str` variable can be NULL here.

```suggestion
	DEBUG(DEBUG_XFRD,2, (LOG_INFO, "CN from cert %s does not match acl", (common_name_str?common_name_str:"(null)")));
```

options.c
```diff
@@ -2158,6 +2197,146 @@ acl_addr_match_range_v6(uint32_t* minval, uint32_t* x, uint32_t* maxval, size_t
...
+			/* Make sure there isn't an embedded NUL character in the DNS name */
+			if ((size_t)ASN1_STRING_length(current_name->d.dNSName) != strlen(dns_name)) {
+				DEBUG(DEBUG_XFRD,2, (LOG_WARNING, "SAN Malformed certificate"));
+				break;
+			}
+			if (dns_name != NULL) {
+				/* certificate SAN DNS match */
+				if (strncmp(dns_name, acl_cert_cn, strlen(acl_cert_cn))==0) {
+					DEBUG(DEBUG_XFRD,2, (LOG_INFO, "SAN %s matches acl", dns_name));
+					*cert_cn = strndup(dns_name, strlen(dns_name));
```
wcawijngaards: Since it is possible to perform strlen, it is zero terminated and with the entire length wanted, ordinary strdup can be used, which may avoid a dependency of NSD on the function from a system library.

```suggestion
					*cert_cn = strdup(dns_name);
```

options.c
```diff
@@ -2158,6 +2197,146 @@ acl_addr_match_range_v6(uint32_t* minval, uint32_t* x, uint32_t* maxval, size_t
...
+	common_name_str = (char *) ASN1_STRING_get0_data(common_name_asn1);
+	/* Make sure there isn't an embedded NUL character in the CN */
+	if ((size_t)ASN1_STRING_length(common_name_asn1) != strlen(common_name_str)) {
+		DEBUG(DEBUG_XFRD,2, (LOG_WARNING, "CN Malformed certificate"));
+		X509_free(client_cert);
+		return 0;
+	}
+	if (common_name_str != NULL) {
+		/* store it for logging */
+		*cert_cn = strndup(common_name_str, strlen(common_name_str));
```
wcawijngaards: Since it is possible to perform strlen, it is zero terminated and with the entire length wanted, ordinary strdup can be used, which may avoid a dependency of NSD on the function from a system library.

```suggestion
		*cert_cn = strdup(common_name_str);
```

options.c
```diff
@@ -2158,6 +2197,146 @@ acl_addr_match_range_v6(uint32_t* minval, uint32_t* x, uint32_t* maxval, size_t
...
+			if (dns_name == NULL)
+				continue;
+			/* Make sure there isn't an embedded NUL character in the DNS name */
+			if ((size_t)ASN1_STRING_length(current_name->d.dNSName) != strlen(dns_name)) {
+				DEBUG(DEBUG_XFRD,2, (LOG_WARNING, "SAN Malformed certificate"));
+				break;
+			}
+			if (dns_name != NULL) {
+				/* certificate SAN DNS match */
+				if (strncmp(dns_name, acl_cert_cn, strlen(acl_cert_cn))==0) {
```
wcawijngaards: The references use case insensitive compare, of the entire string. If only a substring is compared, an acl that allows for example `www.example.com` would allow `www.example.com.example.net` certificates. So I think this is what is meant, unless there is a reason for wanting to do this?

```suggestion
				if (strcasecmp(dns_name, acl_cert_cn)==0) {
```

options.c
```diff
@@ -2158,6 +2197,146 @@ acl_addr_match_range_v6(uint32_t* minval, uint32_t* x, uint32_t* maxval, size_t
...
+	if ((size_t)ASN1_STRING_length(common_name_asn1) != strlen(common_name_str)) {
+		DEBUG(DEBUG_XFRD,2, (LOG_WARNING, "CN Malformed certificate"));
+		X509_free(client_cert);
+		return 0;
+	}
+	if (common_name_str != NULL) {
+		/* store it for logging */
+		*cert_cn = strndup(common_name_str, strlen(common_name_str));
+		/* certificate CN match */
+		if (strncmp(common_name_str, acl_cert_cn, strlen(acl_cert_cn))==0) {
```
wcawijngaards: The same as the other strncmp to strcasecmp comment. The references use case insensitive compare, of the entire string. If only a substring is compared, an acl that allows for example `www.example.com` would allow `www.example.com.example.net` certificates. So I think this is what is meant, unless there is a reason for wanting to do this?

```suggestion
		if (strcasecmp(common_name_str, acl_cert_cn)==0) {
```

options.c
```diff
@@ -1955,6 +1972,28 @@ acl_check_incoming(struct acl_options* acl, struct query* q,
...
+#ifdef HAVE_SSL
+		/* we are in a acl with tls_auth */
+		if (acl->tls_auth_name && q->tls_auth) {
+			/* we have auth_domain_name in tls_auth */
+			if (acl->tls_auth_options && acl->tls_auth_options->auth_domain_name) {
+				q->cert_cn = NULL;
+				if (!acl_tls_hostname_matches(q->tls_auth, acl->tls_auth_options->auth_domain_name, &(q->cert_cn))) {
+					VERBOSITY(3, (LOG_WARNING,
+							"client cert %s does not match %s %s",
+							q->cert_cn, acl->tls_auth_name, acl->tls_auth_options->auth_domain_name));
```
wcawijngaards: The `q->cert_cn` parameter can be NULL after the `acl_tls_hostname_matches` function call.

```suggestion
							(q->cert_cn?q->cert_cn:"(null)"), acl->tls_auth_name, acl->tls_auth_options->auth_domain_name));
```

server.c
```diff
@@ -2138,13 +2142,14 @@ server_tls_ctx_setup(char* key, char* pem, char* verifypem)
 	}
 	listen_sslctx_setup_2(ctx);
 	if(verifypem && verifypem[0]) {
-		if(!SSL_CTX_load_verify_locations(ctx, verifypem, NULL)) {
+		if(!SSL_CTX_load_verify_locations(ctx, verifypem, NULL) ||
+		   !SSL_CTX_set_default_verify_paths(ctx)) {
```
wcawijngaards: The idea here is not to make changes that change the matter for users that do not use this new feature. The addition of these certificates may not be wanted for them.

```suggestion
		if(!SSL_CTX_load_verify_locations(ctx, verifypem, NULL)) {
```

server.c
```diff
@@ -2138,13 +2142,14 @@ server_tls_ctx_setup(char* key, char* pem, char* verifypem)
...
 			log_crypto_err("Error in SSL_CTX verify locations");
 			SSL_CTX_free(ctx);
 			return NULL;
 		}
 		SSL_CTX_set_client_CA_list(ctx, SSL_load_client_CA_file(verifypem));
-		SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, NULL);
+		SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT, NULL);
```
wcawijngaards, line 2151: The idea is that code does not change for users that do not use the new feature. This new option, I think it is new from more recent library versions, may be very nice, but the existing code already checks for a peer certificate, in the code in options.c that checks the acl. So I think it is not actually necessary to change this, for other code paths.

```suggestion
		SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, NULL);
```

server.c
```diff
@@ -5258,7 +5315,11 @@ handle_tcp_accept(int fd, short event, void* arg)
 	tcp_data->query_count = 0;
 #ifdef HAVE_SSL
 	tcp_data->shake_state = tls_hs_none;
-	tcp_data->tls = NULL;
+	/* initialize both incase of dangling pointers */
+	if (data->tls_accept || data->tls_auth_accept) {
+		tcp_data->tls = NULL;
+		tcp_data->tls_auth = NULL;
+	}
```
wcawijngaards: I think it could always initialize the member variables to NULL? That may be more prudent. Is it correct that this is possible? If so the code suggestion:
```suggestion
	/* initialize both incase of dangling pointers */
	tcp_data->tls = NULL;
	tcp_data->tls_auth = NULL;
```

wcawijngaards (changes requested):
Thank you for the updates for the patch. It is very nice to have good mutual authentication. During review I found some issues that are serious enough to request changes before approval; a memory leak and a question on name comparison. The reason I am asking though, is also to make sure these changes are improvements, perhaps there is good reason for doing it this way. Maybe that code is there on purpose?

The memory leak and name comparisons have suggested code changes, and I would approve the patch if they are applied; and if you also think these are the correct way of handling the certificate names. Comparing only a substring of the name was I think not desirable, so I suggested strcasecmp that compares the entire name, and also case insensitively, since the examples from the reference do that too.

The suggestion is to remove `SSL_CTX_set_default_verify_paths`, since the extra certificates could be a surprise for users that only want to use their own .pem file. And then suddenly public certificate authorities can allow access with certificates. I guess a separate change could add it, and also the newer fail if no peer cert option. A separate patch is not needed, right now. Although that is nicer for not changing the code path when this feature is not used, the question is if the suggested changes are alright for this patch.

server.c
bilias: I believe this is ok. Previous behaviour never reached there.

https://github.com/NLnetLabs/nsd/blob/a91e67c553b2933072d3238a098e8ef65379b143/server.c#L2144-L2155
is called from
https://github.com/NLnetLabs/nsd/blob/a91e67c553b2933072d3238a098e8ef65379b143/server.c#L2059
https://github.com/NLnetLabs/nsd/blob/a91e67c553b2933072d3238a098e8ef65379b143/server.c#L2176
https://github.com/NLnetLabs/nsd/blob/a91e67c553b2933072d3238a098e8ef65379b143/server.c#L2158
https://github.com/NLnetLabs/nsd/blob/a91e67c553b2933072d3238a098e8ef65379b143/nsd.c#L1578-L1593

So previous usage was line 1582 of nsd.h where verypem was always null, thus SSL_CTX_load_verify_locations() was never called.

server.c
bilias: Same as before.

server.c
bilias: This is tricky and we need to be carefull here.
I has a long time bug here, which I was chasing for a long time while developing the patch.

I had random crashes untill I did NULL on both tls_auth and tls.
I had cases where I was in tls and tls_auth was not null, that's why I initialized both of them.

I believe your change is fine, its a more general rule. To my understadning handle_tcp_accept() is unique for each connection.

server.c
wcawijngaards: But not from remote.c:272, where server_tls_ctx_setup is also called.

bilias: you are correct

options.c
bilias: Are you sure about this? I've never used double pointer before so I can't really tell if free is needed here.
If you're ok then I will commit it.

server.c
bilias: According to manual page
```
      SSL_VERIFY_FAIL_IF_NO_PEER_CERT
           Server  mode:  if  the  client  did  not  return  a certificate, the TLS/SSL handshake is immediately 
           terminated with a"handshake failure" alert.  This flag must be used together with SSL_VERIFY_PEER.
```
My understanding is that SSL_VERIFY_PEER is not enough for verification.

wcawijngaards: That is true, but the code performs a specific check for no peer certificate already, so the flag is not really necessary, and I was being very cautious. But otherwise this is okay.

options.c
wcawijngaards: Just double checked the code paths, I am certain. The value is set to a strdup of a string or left at NULL, and this then needs to be freed.

wcawijngaards (approved):
The changes are looking good! Is it alright if I merge this now, or are there more edits you want to make?

server.c
bilias: We leave this as it is?
I believe it does not hurt to initialize every time, apart from the extra cycles when we're not in a TLS listener.

wcawijngaards: The commit already changes the code https://github.com/NLnetLabs/nsd/pull/337/commits/a19841c18a50c8368ad8aaf326fcc6ee785f1852 . So I think it is fine, I wanted to make sure it was okay, which edit is the one you think best here.

bilias:
> The changes are looking good! Is it alright if I merge this now, or are there more edits you want to make?

I'm OK.
I wanted to also implement regression tests for these changes, but this can come on a new PR
and probably after a few weeks.

Thanks for looking into this :)

PR #341, opened 2024-06-20
Fix allow-query wording in nsd.conf.5.in

realPyR3X:
Fix allow-query wording with the implied intent.


wcawijngaards (approved):
Grammar was okay, but double words are removed. Otherwise it is fine to change.

wcawijngaards:
Thanks for the fix! It makes the documentation better, it is merged into the code repository.

Issue #347, opened 2024-07-05
Adjust verbosity for TLS (+TCP) to be 5


jaredmauch:
As we are seeing an increase in TCP queries this creates excessive logging when we would not want those events logged at lower levels while we would want to see other logs.  Move the TLS+TCP to be verbosity level 5

wcawijngaards:
Verbosity level 3 is a little low for that output. This changes the output of `TLS handshake: connection closed prematurely` and `TLS handshake succeeded.`. If there are more messages that need to change verbosity level, those would need to be changed too.

PR #348, opened 2024-07-05
Move TLS logging to verbosity level 5

jaredmauch:
Patch for https://github.com/NLnetLabs/nsd/issues/347


wcawijngaards (approved):
This change is fine. It also changed the verbosity level of the 'setup TLS' message.

wcawijngaards:
Thanks for the change patch! The other commits for #347 change a couple other log messages for TCP and TLS to verbosity level 5 as well.

PR #349, opened 2024-07-06
log file name before loading

jaredmauch:
in the event there is a fatal error inside zonec_read provide ability to log the zone and file names prior to entry


jaredmauch:
@wcawijngaards do you mind including this?

wcawijngaards (approved):
The code looks fine for additional information.

wcawijngaards:
Certainly, it is a nice feature. I included it, and also changed it to verbosity 5; I think that may be a better verbosity level for it. Most production usage is with smaller values.

PR #350, opened 2024-07-11
Use numeric values when serializing WKS RRs

k0ekk0ek:
See description in `rdata.c`


k0ekk0ek:
I'll also update the description in src/generic/wks.h (which is already quite extensive).

wcawijngaards (approved):
Printing the record in a compatible format is a good idea. The code looks fine.

PR #351, opened 2024-07-12
Use MAKE variable rather than make command

k0ekk0ek:
Calling `make` directly resulted in the wrong make being executed for simdzone. This fixes that.


wcawijngaards (approved):
Makes the make clean target better. The patch looks nice.

PR #352, opened 2024-07-15
contrib: add OpenRC service script, config file, and tmpfiles entry

orlitzky:
1. Add an OpenRC service script and config file under `contrib/`. The service script is integrated with the build system so that the correct paths are obtained from `./configure` and do not need to be hard-coded. (This requires >=autoconf-2.70.) Tested on Gentoo, but it should work on other distros that use OpenRC as well.

2. Add a tmpfiles entry under `contrib/` for both systemd and OpenRC. If installed, this will create at a boot a temporary "nsd" directory writable by the nsd user. This makes it easier to use a local control socket because the user no longer has to worry about creating the directory where the socket will live.

3. Tweak the default config to show an example of using a local socket in the tmpfiles location.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@
 /udb-inspect
 /xfr-inspect
 /contrib/nsd.openrc
+/contrib/nsd-tmpfiles.conf
 /dnstap/dnstap.pb-c.c
 /dnstap/dnstap.pb-c.h
 /dnstap/dnstap_config.h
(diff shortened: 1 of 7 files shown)


wcawijngaards (approved):
The config looks nice and deals with the pidfile neatly. The contribution is appreciated!

orlitzky:
Thank you!

PR #360, opened 2024-07-26
Replace Cirrus CI with GitHub Actions

k0ekk0ek:
This PR replaces Cirrus CI with GitHub Actions. For some reason the macOS runners behave weird when trying to figure out if stdin/stdout/stderr are closed, so I disabled some tests. This also makes Coverity Scan work again.


dname.h
```diff
@@ -182,7 +182,7 @@ dname_label(const dname_type *dname, uint8_t label)
  *
  * Pre: left != NULL && right != NULL
  */
```
wcawijngaards, line 185: Maybe noting the type in the comment is nice.
```suggestion
 * left and right are dname_type*.
 */
```

wcawijngaards (approved):
The move to github actions looks good. It disables use of cirrus ci.

dname.h
k0ekk0ek: Fixed

PR #362, opened 2024-07-29
Implement fallback to ASN1_STRING_data

k0ekk0ek:
The version used of OpenSSL used by the opencsw project is too old and does not offer `ASN1_STRING_get0_data`. Implement falling back to `ANS1_STRING_data` instead. Also fixes `sk_GENERAL_NAME_pop_free` not being called if the subject alternative name matched previously. Apart from that, I also removed duplication of the common name, I'm not sure it adds useful debugging info.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -359,11 +359,11 @@ AC_DEFUN([CHECK_SSL], [
         ])
     if test x_$withval != x_no; then
         AC_MSG_CHECKING(for SSL)
-	if test -n "$withval"; then
-        	dnl look for openssl install with different version, eg.
-                dnl in /usr/include/openssl11/openssl/ssl.h
-                dnl and /usr/lib64/openssl11/libssl.so
-                dnl with the --with-ssl=/usr/include/openssl11
+        if test -n "$withval"; then
+            dnl look for openssl install with different version, eg.
+            dnl in /usr/include/openssl11/openssl/ssl.h
+            dnl and /usr/lib64/openssl11/libssl.so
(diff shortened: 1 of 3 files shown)


wcawijngaards (approved):
The code looks good. Avoids useless allocations, and it still has debug output.

k0ekk0ek:
@bilias, this slightly updates your work on #337. If you can spare the time, can you verify everything still works for you? Thanks in advance.

k0ekk0ek:
The person reporting the error via email confirmed NSD compiles on Solaris again with both GCC and the Sun compiler, so we're good there.

bilias:
@k0ekk0ek 

Just tested https://github.com/k0ekk0ek/nsd/tree/test-for-asn1-string-get0-data
and it passed my tests both for allow and deny access depending on certificate (CN and SAN).

Code looks good to me and I like the simplification of the code after your changes.

k0ekk0ek:
Thanks for testing @bilias!

bilias:
@k0ekk0ek 
My only concern is the removal of those DEBUG statements, especially after reading Andreas mail and his last two mentioned notes.

Trying to find out why TLS-auth does not work can be tricky some times, especially with the default output log of openssl which sucks (in my opinion). That's why I did all that fuzz and added all those extra DEBUG.

Maybe some errors deserve a (better) VERBOSITY log.

k0ekk0ek:
@bilias, I've created a separate issue for it (#369) and we can do it for the next release. The resource leak and build issue on Solaris have been fixed, so I'd like to focus on 4.10.1 first.

PR #366, opened 2024-07-30
Be consistent in naming primary/secondary zones

jpmens:
BIND's terminology has also changed and the terms have been deprecated (also in their configuration) for the newer primary / secondary. As this man page also uses the terms, I think we should be consistent throughout.


k0ekk0ek:
Thanks @jpmens! These changes look good :+1:

PR #368, opened 2024-07-31
Various fixes to the manpages `nsd-checkzone(8)` and `nsd.conf(5)`

cschug:
In `nsd-checkzone(8)`:

- Document that `nsd-checkzone(8)` can read zones from stdin (again) as per NLnetLabs/simdzone#205

In `nsd.conf(5)`:

- Remove bogus references to the `-n` command-line option
- Reference command-line option `-V` which corresponds to the `verbosity` setting
- Simplify the extraction of the OCSP responder URL in the example given at the `tls-service-ocsp` setting as any non-ancient `openssl x509` has a support for that
- Typo fix and reference to `nsd-checkzone(8)` in the "SEE ALSO" section as it is mentioned in the `create-ixfr` stanza

### DIFF
--- a/nsd-checkzone.8.in
+++ b/nsd-checkzone.8.in
@@ -24,7 +24,8 @@ Print usage help information and exit.
 The name of the zone to check, eg. "example.com".
 .TP
 .I zonefile
-The file to read, eg. "zones/example.com.zone.signed".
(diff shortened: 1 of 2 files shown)


k0ekk0ek (approved):

k0ekk0ek:
Thanks @cschug! Changes lgtm. I'll merge once the checkboxes turn green.

PR #370, opened 2024-08-02
Update badges

k0ekk0ek:
What it says on the tin. Replaces the Cirrus CI badge with a GitHub Actions one and adds a Coverity Scan badge.


wcawijngaards (approved):
The cirrus badge would not work any more. Go ahead and add badges.

PR #371, opened 2024-08-05
Import documentation from nsd-manual repository

k0ekk0ek:
I thought I'd finally move the documentation to the main repository because it's really outdated. There's a difference between what's in the `nsd-manual` repository and this PR and that's making use of `mandoc` to convert the manpages we currently have to html and including that in the `.rst` file. The output looks good IMHO and allows us to keep the rest as is (not quite what was discussed, but a good tradeoff?).


k0ekk0ek:
There, all done(?) A preview can be found [here](https://nsd-manual-in-repo-test.readthedocs.io/en/latest/index.html).

PR #373, opened 2024-08-09
Add support for reloading configuration on SIGHUP

### LINKED ISSUE #323: Add zones without nsd-control
If I change one of my zones, I can reload it using  `systemctl reload nsd`. This makes use of systemd's  [ExecReload](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#ExecReload=), which can be overridden but typically just sends `SIGHUP` to the daemon's PID. This is especially nice because the unprivileged `nsd` user is allowed to send `SIGHUP` to his own process, so changing zones does not require root.

If I _add_ a zone, however, I have to change `nsd.conf` or one of its include files, and I've noticed that `SIGHUP` will not cause nsd to "reload" the new zone. Instead, nsd-control refers to that as "reconfig". If nsd-control is not enabled, a full restart of the daemon is therefore required to add a zone, and that requires root.

Is there some way around this? The description of the systemd feature says that it "triggers a configuration reload," and it would be nice if there was a way to force a reconfig instead of just a zone reload using that mechanism. Maybe a reconfig could be triggered with `SIGUSR2`?

In short, the problem I'm trying to solve is to be able to add new zones locally without root privileges and without nsd-control.

k0ekk0ek:
Fixes #323.

### DIFF
--- a/configlexer.lex
+++ b/configlexer.lex
@@ -267,6 +267,7 @@ rrl-ipv4-prefix-length{COLON}	{ LEXOUT(("v(%s) ", yytext)); return VAR_RRL_IPV4_
 rrl-ipv6-prefix-length{COLON}	{ LEXOUT(("v(%s) ", yytext)); return VAR_RRL_IPV6_PREFIX_LENGTH;}
 rrl-whitelist-ratelimit{COLON}	{ LEXOUT(("v(%s) ", yytext)); return VAR_RRL_WHITELIST_RATELIMIT;}
 rrl-whitelist{COLON}	{ LEXOUT(("v(%s) ", yytext)); return VAR_RRL_WHITELIST;}
+reload-config{COLON}	{ LEXOUT(("v(%s) ", yytext)); return VAR_RELOAD_CONFIG; }
 zonefiles-check{COLON}	{ LEXOUT(("v(%s) ", yytext)); return VAR_ZONEFILES_CHECK;}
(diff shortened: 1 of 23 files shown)


orlitzky:
Thanks for this! Since SIGHUP/reload are mentioned in  https://github.com/NLnetLabs/nsd/blob/master/doc/README#L607, it might help to compare the two? If I understand correctly, `nsd-control reload` always reloads everything, but SIGHUP only reloads the zones unless the new option is enabled.

For that matter, it might also make sense to mention the new option in https://github.com/NLnetLabs/nsd/blob/master/nsd.8.in#L209?

wcawijngaards (approved):
Looks good, the option could also be coded in config_print_zone() in nsd-checkconf.c:285, and config_test_print_server() in nsd-checkconf.c:605 and then see if tpkg/checkconf.tdir needs changes too. Nice to have backwards compatible behaviour with the new option to also read in new zones, drop old zones, and change tsig keys, from the config file.

k0ekk0ek:
Thanks @wcawijngaards. Forgot about checkconf. I'll add it (and the manpage updates suggested by @orlitzky).

k0ekk0ek:
@orlitzky
>  If I understand correctly, nsd-control reload always reloads everything, but SIGHUP only reloads the zones unless the new option is enabled.

No, I hooked up the new option to logic that was already implemented for `nsd-control reconfig`. Updated the man page.

PR #376, opened 2024-08-22
Point the user towards tcpdump for logging individual queries

hills:
The lack of logging is explained, but appears to assume the user is interested in aggregate log. For simple debugging there is no equivalent to "log-queries" (or "log-replies") in unbound(8).

If the method described in [*] is the recommended one then it should be documented as such and also saves time.

It's not clear to me what the difference between doc/*.rst and doc/README is. This patch assumes README is already out of date.

[*] https://lists.nlnetlabs.nl/pipermail/nsd-users/2020-June/002880.html


wcawijngaards (approved):
Additional documentation is useful. Thank you for the contribution!

PR #377, opened 2024-08-23
Track $INCLUDE dependencies

### LINKED ISSUE #57: Doesn't reload zone when included files change
When I make a change to a file `$INCLUDE`d in a zone file, `nsd` cowardly refuses to reload the zone, claiming that the zone file hasn't been modified on disk. It hasn't, but `nsd` should be checking included files as well, and reload if any files in the include tree have been modified.

k0ekk0ek:
What it says on the tin. Fixes #57.

### DIFF
--- a/dbaccess.c
+++ b/dbaccess.c
@@ -75,6 +75,8 @@ namedb_zone_create(namedb_type* db, const dname_type* dname,
 	zone->opts = zo;
 	zone->ixfr = NULL;
 	zone->filename = NULL;
+	zone->includes.count = 0;
+	zone->includes.paths = NULL;
 	zone->logstr = NULL;
 	zone->mtime.tv_sec = 0;
 	zone->mtime.tv_nsec = 0;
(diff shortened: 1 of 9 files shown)


namedb.h
```diff
@@ -139,7 +139,13 @@ struct zone
 #endif
 	struct zone_options* opts;
 	struct zone_ixfr* ixfr;
-	char*        filename; /* set if read from file, which file */
+	char *filename; /* set if read from file, which files */
+	/* give a description here of why we maintain a "list" of file
+	 * names */
```
wcawijngaards: The comment seems to be todo.

dbaccess.c
```diff
@@ -89,6 +91,35 @@ namedb_zone_create(namedb_type* db, const dname_type* dname,
 	return zone;
 }
 
+void
+namedb_zone_free_filenames(namedb_type *db, zone_type* zone)
+{
+	assert(!zone->includes.paths || !zone->filename);
```
wcawijngaards: I do not get this assertion. It seems that the zone could have a filename, and that while parsing includes.paths can get filled with include files. That makes them both not NULL and that would fail this assertion.

dbaccess.c
```diff
@@ -235,13 +264,28 @@ namedb_read_zonefile(struct nsd* nsd, struct zone* zone, udb_base* taskudb,
 
 		/* if zone_fname, then the file was acquired from reading it,
 		 * and see if filename changed or mtime newer to read it */
-		} else if(zone_fname && strcmp(zone_fname, fname) == 0 &&
-		   timespec_compare(&zone_mtime, &mtime) == 0) {
-			VERBOSITY(3, (LOG_INFO, "zonefile %s is not modified",
-				fname));
-			return;
+		} else if(zone_fname && strcmp(zone_fname, fname) >= 0 &&
```
wcawijngaards: The name comparison is wrong, it seems.
```suggestion
		} else if(zone_fname && strcmp(zone_fname, fname) == 0 &&
```

wcawijngaards (approved):
Nice to keep track of include files. It looks okay.

dbaccess.c
k0ekk0ek: Updated.

dbaccess.c
k0ekk0ek: Fixed.

namedb.h
k0ekk0ek: Fixed.

Issue #196, opened 2021-10-06
Default location of "cookie-secret-file"


anandb-ripencc:
The nsd.conf man page says that the default location of the cookie secret file is /etc/nsd/nsd_cookiesecrets.txt. However, if one tries to add a cookie using nsd-control, this is what happens:

```
# nsd-control add_cookie_secret ae44cc3c8ceaaa64f5811e9a55442ab6
unable to open cookie secret file /etc/nsd/nsd_cookiesecrets.txt: Permission deniederror: writing to cookie secret file: "/etc/nsd/nsd_cookiesecrets.txt"
```

The reason for this is that `/etc/nsd` is usually not writable by anyone other than "root", but the NSD server normally runs as a non-root user. This is generally true of most daemons on unix-like systems. If a running daemon wants to create some kind of state file, such a file should normally go into a location such as `/var/lib/nsd`, which *is* writeble by the "nsd" user. In fact, all of NSD's other state files (nsd.db, xfrd.state, dynamically added zone list, etc) go here. So I find it surprising that the dynamic cookie secret file's default location is `/etc/nsd`. It does not follow best practices for unix-like systems.

I then tried to see if I could change this default at compile time, but there's no `configure` option for this either (whereas there are options for specifying default locations for the log file, database file, state file, zone list file, etc). I consider this a bug, and I think that you should change the default to point to the same place as NSD's other run-time state files.

anandb-ripencc:
This issue is still open, and a recent discussion of cookies and run-time reconfiguration brought me back here. While this bug needs to be fixed, I also feel like cookie secrets are something that should be modifiable at run-time with a reconfig. BIND makes this very easy, by allowing multiple secrets to be defined. The first is used as the "active" one, while all others after it are used only for verification. Rolling the secret is made easy, by just changing the order of the secret in `named.conf` and reloading the configuration.

For NSD, instead of using a cookie file, I can envision having secrets in `nsd.conf`, similar to BIND, with the first one used as the active secret. Since `secret` is already an option, you could introduce `staging-secret` as an additional option, to specify one or more incoming secrets. The operator can just swap `secret` and `staging-secret` around, and reload the configuration.

This is of course dependent on the ability to reconfigure NSD on the fly, as I have suggested in #374.

wtoorop:
[...]

PR #378, opened 2024-08-28
Updated cookie secrets management

wtoorop:
This PR is in response to issue #196 (and thus also of interest to @anandb-ripencc  )
- The cookie secret file's default location can be set with `--with-cookiesecretsfile=path` with `configure`
- The default cookie secrets file is changed to `{dbdir}/cookiesecrets.txt`
- When no cookie secrets file location is set with the `cookie-secret-file` option in the config file, and the new default location does not exist, cookie secrets will be read from the previous default location `{configdir}/nsd_cookiesecrets.txt`
- A staging cookie secret can also be provided in the config file with the `cookie-staging-secret` option.
- Explicitly configured cookie secrets in config file (with the `cookie-secret` and `cookie-staging-secret` option) take precedence over the ones from the cookie secret file
- All values of DNS cookie related settings from config file (`answer-cookie`, `cookie-secret`, `cookie-staging-secret` and `cookie-secret-file`) will be reevaluated and effectuated after `nsd-control reconfig`


wtoorop:
It's ready for review (and merge if accepted). One thing to consider/discuss is that it has one thing that is not backwards compatible, which is that cookie secrets from config file now are preferred over cookies from the cookie secrets file. Before this was the other way around. But, this feels more logical to me as zones in the configuration file are also preferred over zones in the zone list file. I believe there will be a negligible number of people that have cookie secrets in both the config file and in the cookie secrets file and would depend on the earlier preference to pick the secrets file.

anandb-ripencc:
Looks reasonable to me. I agree that despite the backwards-incompatible change, it is unlikely to affect anyone, because no-one would define cookies in both places.

difffile.c
```diff
@@ -1988,53 +1969,13 @@ task_process_del_key(struct nsd* nsd, struct task_list_d* task)
...
-		return;
-	}
-	activate_cookie_secret(nsd);
+task_process_cookies(struct nsd* nsd, struct task_list_d* task) {
+	DEBUG(DEBUG_IPC, 1, (LOG_INFO, "cookies task answer: %s, count: %d",
+		task->newserial ? "yes" : "no", (int)task->yesno));
+	
+	nsd->do_answer_cookie = (int) task->newserial;
+	nsd->cookie_count = (size_t) task->yesno;
+	memmove(nsd->cookie_secrets, task->zname, sizeof(nsd->cookie_secrets));
```
wcawijngaards, line 1978: The function `task_process_add_cookie_secret` that was deleted used `explicit_bzero`, to wipe the secrets after use. The change adds that to this function.

```suggestion
	memmove(nsd->cookie_secrets, task->zname, sizeof(nsd->cookie_secrets));
	explicit_bzero(task->zname, sizeof(nsd->cookie_secrets));
```

remote.c
```diff
@@ -2366,65 +2430,75 @@ cookie_secret_file_dump(RES* ssl, nsd_type const* nsd) {
...
-	(void)cookie_secret_file_dump(ssl, nsd);
-	task_new_activate_cookie_secret(xfrd->nsd->task[xfrd->nsd->mytask],
-	    xfrd->last_task);
-	xfrd_set_reload_now(xfrd);
-	send_ok(ssl);
+	if(!cookie_secret_file_dump_and_reload(ssl, nsd)) {
+		memcpy( nsd->cookie_secrets, backup_cookie_secrets
+		      , sizeof(cookie_secrets_type));
+		nsd->cookie_count = backup_cookie_count;
+	}
```
wcawijngaards, line 2469: The `backup_cookie_secrets` variable can be erased after the function is done.

```suggestion
	}
	explicit_bzero(backup_cookie_secrets, sizeof(cookie_secrets_type));
```

remote.c
```diff
@@ -2366,65 +2430,75 @@ cookie_secret_file_dump(RES* ssl, nsd_type const* nsd) {
...
-	(void)cookie_secret_file_dump(ssl, nsd);
-	task_new_drop_cookie_secret(xfrd->nsd->task[xfrd->nsd->mytask],
-	    xfrd->last_task);
-	xfrd_set_reload_now(xfrd);
-	send_ok(ssl);
+	if(!cookie_secret_file_dump_and_reload(ssl, nsd)) {
+		memcpy( nsd->cookie_secrets, backup_cookie_secrets
+		      , sizeof(cookie_secrets_type));
+		nsd->cookie_count = backup_cookie_count;
+	}
```
wcawijngaards, line 2495: The `backup_cookie_secrets` variable can be erased after the function is done.

```suggestion
	}
	explicit_bzero(backup_cookie_secrets, sizeof(cookie_secrets_type));
```

remote.c
```diff
@@ -2443,27 +2517,24 @@ do_add_cookie_secret(RES* ssl, xfrd_state_type* xrfd, char* arg) {
...
-	xfrd_set_reload_now(xfrd);
-	send_ok(ssl);
+	if(!cookie_secret_file_dump_and_reload(ssl, nsd)) {
+		explicit_bzero(arg, strlen(arg));
+		(void)ssl_printf(ssl, "error: writing to cookie secret file: \"%s\"\n"
+		                , cookie_secret_file(nsd->options));
+		memcpy( nsd->cookie_secrets, backup_cookie_secrets
+		      , sizeof(cookie_secrets_type));
+		nsd->cookie_count = backup_cookie_count;
+	}
```
wcawijngaards, line 2544: The `backup_cookie_secrets` variable can be erased after the function is done.

```suggestion
	}
	explicit_bzero(backup_cookie_secrets, sizeof(cookie_secrets_type));
```

util.c
```diff
@@ -1145,3 +1149,110 @@ void drop_cookie_secret(struct nsd* nsd)
...
+		secret_len = strlen(secret);
+		if( secret_len == 0 ) { break; }
+		assert( secret_len <= sizeof(secret) );
+		secret_len = secret[secret_len - 1] == '\n' ? secret_len - 1 : secret_len;
+		if( secret_len != NSD_COOKIE_SECRET_SIZE * 2 ) {
+			fclose(f);
+			log_msg( LOG_ERR
+			       , "error parsing cookie secret file \"%s\""
+			       , fn);
+			return 0;
```
wcawijngaards: The `cookie_secrets` and `secret` variables can be erased when the function is done.

```suggestion
			explicit_bzero(cookie_secrets, sizeof(cookie_secrets));
			explicit_bzero(secret, sizeof(secret));
			return 0;
```

util.c
```diff
@@ -1145,3 +1149,110 @@ void drop_cookie_secret(struct nsd* nsd)
...
+		/* needed for `hex_pton`; stripping potential `\n` */
+		secret[secret_len] = '\0';
+		decoded_len = hex_pton(secret, cookie_secrets[count].cookie_secret,
+		                       NSD_COOKIE_SECRET_SIZE);
+		if( decoded_len != NSD_COOKIE_SECRET_SIZE ) {
+			fclose(f);
+			log_msg( LOG_ERR
+			       , "error parsing cookie secret file \"%s\""
+			       , fn);
+			return 0;
```
wcawijngaards: The `cookie_secrets` and `secret` variables can be erased when the function is done.

```suggestion
			explicit_bzero(cookie_secrets, sizeof(cookie_secrets));
			explicit_bzero(secret, sizeof(secret));
			return 0;
```

util.c
```diff
@@ -1145,3 +1149,110 @@ void drop_cookie_secret(struct nsd* nsd)
...
+	}
+	fclose(f);
+	if(count && nsd->cookie_secrets_source <= COOKIE_SECRETS_FROM_FILE) {
+		nsd->cookie_count = count;
+		memcpy(nsd->cookie_secrets, cookie_secrets, sizeof(cookie_secrets));
+		nsd->cookie_secrets_source = COOKIE_SECRETS_FROM_FILE;
+		region_str_replace(  nsd->region
+		                  , &nsd->cookie_secrets_filename, fn );
+	}
+	return 1;
```
wcawijngaards: The `cookie_secrets` variable can be erased when the function is done.

```suggestion
	explicit_bzero(cookie_secrets, sizeof(cookie_secrets));
	return 1;
```

util.c
```diff
@@ -1145,3 +1149,110 @@ void drop_cookie_secret(struct nsd* nsd)
...
+		secret[secret_len] = '\0';
+		decoded_len = hex_pton(secret, cookie_secrets[count].cookie_secret,
+		                       NSD_COOKIE_SECRET_SIZE);
+		if( decoded_len != NSD_COOKIE_SECRET_SIZE ) {
+			fclose(f);
+			log_msg( LOG_ERR
+			       , "error parsing cookie secret file \"%s\""
+			       , fn);
+			return 0;
+		}
```
wcawijngaards, line 1272: The `secret` variable can be erased after use.

```suggestion
		explicit_bzero(secret, sizeof(secret));
		}
```

wcawijngaards (approved):
The code looks all right. I saw some places where the cookie secrets could possibly be wiped after temporary variables are done being used, maybe that is nice to have.

remote.c
```diff
@@ -2473,7 +2547,23 @@ do_print_cookie_secrets(RES* ssl, xfrd_state_type* xrfd, char* arg) {
...
+	case COOKIE_SECRETS_FROM_FILE:
+		ssl_printf( ssl, "source : \"%s\"\n"
+		          , nsd->cookie_secrets_filename);
+		break;
+	case COOKIE_SECRETS_FROM_CONFIG:
+		ssl_printf(ssl, "source : configuration\n");
+		break;
+	default:
+		ssl_printf(ssl, "source : unknown\n");
+		break;
```
k0ekk0ek, line 2570: All possibilities are covered, so this is a panic situation(?)

util.c
```diff
@@ -1145,3 +1149,116 @@ void drop_cookie_secret(struct nsd* nsd)
...
+
+	/* a non-existing cookie file is not an error */
+	else if(errno != ENOENT) {
+		log_msg( LOG_ERR
+		       , "error reading cookie secret file \"%s\": \"%s\""
+		       , fn, strerror(errno));
+		return 0;
+	}
+	else if(nsd->options->cookie_secret_file != NULL /* explicit name */
+	     ||!(f = fopen((fn = CONFIGDIR"/nsd_cookiesecrets.txt"), "r")))
```
k0ekk0ek: Falling back to `nsd_cookiesecrets.txt` in this function doesn't seem right. I think it's preferred to handle that in the configuration parser (or shortly after). Maybe do something like if `cookie_secret_file` is still `NULL` after parsing, set it to the previous default? That way behavior is more deterministic. Also would avoid the `cookie_secret_file` function, which seems to return `COOKIESECRETSFILE` if the `cookie_secret_file` is `NULL`. So, basically `fn` is only `NULL` if the `cookie_secret_file` is not `NULL` but empty, in which case `NULL` is returned. This seems a very convoluted way of avoiding checking whether the member is `NULL` just after parsing and setting right then(?)

util.c
```diff
@@ -1145,3 +1149,116 @@ void drop_cookie_secret(struct nsd* nsd)
...
+			       , "error parsing cookie secret file \"%s\""
+			       , fn);
+			explicit_bzero(cookie_secrets, sizeof(cookie_secrets));
+			explicit_bzero(secret, sizeof(secret));
+			return 0;
+		explicit_bzero(secret, sizeof(secret));
+		}
+	}
+	fclose(f);
+	if(count && nsd->cookie_secrets_source <= COOKIE_SECRETS_FROM_FILE) {
```
k0ekk0ek: An `enum` is mapped to an integer, but it's better to be explicit. The order might change and this may go unnoticed. Basically, you want the branch to be taken if the secret does not originate from the configuration (`!= COOKIE_SECRETS_FROM_CONFIG`).

k0ekk0ek (approved):
I agree with the changes in general, just some small remarks regarding use of `enums` and one related to falling back to the previous value.

remote.c
wtoorop: I guess yes. Would you prefer a assert here?

wtoorop: Would you prefer an assert?

util.c
wtoorop:
```suggestion
	if(count && nsd->cookie_secrets_source != COOKIE_SECRETS_FROM_FILE) {
```

wtoorop: Done in [5674228](https://github.com/NLnetLabs/nsd/pull/378/commits/5674228f1cd200bbf1e5a645d55e8bbf0b90dd3d)

anandb-ripencc:
I was reviewing the changes to the man pages and documentation, and it needs spelling and syntax fixes. Let me comment on the changes so you can fix them before merging this branch.

wtoorop:
> I was reviewing the changes to the man pages and documentation, and it needs spelling and syntax fixes. Let me comment on the changes so you can fix them before merging this branch.

Thanks! I'm also still addressing the last comment from @k0ekk0ek, which will simplify the code somewhat and also _only does the fallback_ to the old default location for secrets (if there is nothing configured and nothing in the new default location), _when nsd just started_.

util.c
wtoorop: Okay, addressed in [c743742](https://github.com/NLnetLabs/nsd/pull/378/commits/c743742a6ebdc899bb6b65b0f8ff2171662517c1)

wtoorop:
> I was reviewing the changes to the man pages and documentation, and it needs spelling and syntax fixes. Let me comment on the changes so you can fix them before merging this branch.

@anandb-ripencc I am done with review feedback and will merge once you're finished your spelling corrections.

anandb-ripencc:
Actually, just merge it. I will then submit a PR to update the documentation. Please don't make a new release before I've submitted documentation fixes.

k0ekk0ek:
Looks good te me @wtoorop :+1:

wtoorop:
I accidentally committed along changes in simdzone. I'll force push the commit without the simdzone changes, and then merge.

PR #388, opened 2024-10-17
nsd-control: Prefix error and warning messages with "nsd-control: {error,warning}: "

lanodan:
This allows to have less cryptic errors in scripts, specially automated ones.

### DIFF
--- a/nsd-control.c
+++ b/nsd-control.c
@@ -146,7 +146,7 @@ usage()
 /** exit with ssl error */
 static void ssl_err(const char* s)
 {
-	fprintf(stderr, "error: %s\n", s);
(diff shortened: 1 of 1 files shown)


k0ekk0ek:
Hi @lanodan! Thanks for issuing the PR. I feel adding `nsd-control: ` to the output is a bit much and adding `nsd-control:` to the output from the script itself is a better solution(?)

lanodan:
Well… could be done but something like a sed on stderr isn't among the most straightforward things to do in a shell script.

k0ekk0ek:
Sorry, we can't take this patch @lanodan (discussed it with the other devs too). Closing this PR. Don't be shy to send us any fixes or report issues in the future even though we can't take this one.

Issue #387, opened 2024-10-08
SIGSEGV in rbtree_find_less_equal


k0ekk0ek:
As reported on [nsd-users](https://lists.nlnetlabs.nl/pipermail/nsd-users/2024-October/003333.html).

Reproducible seg fault with a DNSSEC signed zone and overlapping config. Running NSD 4.10.1. Here's how to reproduce.

2 zones in `nsd.conf`:

```
zone:
        name:     "foo.com."
        zonefile:     "/zones/foo.com.zone.signed"

zone:
        name:     "bar.foo.com."
        zonefile: "/zones/bar.foo.com.zone"
```

Zone files:
foo.com.zone.signed is DNSSEC signed with a record for a.bar (A record or anything)
bar.foo.com.zone doesn't exist  (but it's in nsd.conf shown above)

Steps:
1) Startup NSD
2) touch foo.com.zone.signed
3) reload NSD

nsd.log will say:
```
[2024-10-02 07:19:58.691] nsd[962739]: info: control cmd:  reload
[2024-10-02 07:19:58.845] nsd[962752]: error: handle_reload_cmd: reload closed cmd channel
[2024-10-02 07:19:58.845] nsd[962752]: warning: Reload process 962740 failed, continuing with old database
```

core dump says SIGSEGV in rbtree_find_less_equal

k0ekk0ek:
The issue can be successfully reproduced using a debug build and the attached configuration.
[nsd.conf](https://github.com/user-attachments/files/17307622/nsd.conf.txt)
[foo.com.zone.signed](https://github.com/user-attachments/files/17307628/foo.com.zone.signed.txt)

PR #389, opened 2024-10-17
Skip over non-NSEC3 zones when looking for dstree

k0ekk0ek:
This fixes the issue mentioned on #387.


wcawijngaards (approved):
This should stop the dstree from being NULL after a zone lookup. NSD should really be able to deal with normal configuration of zones, also zones that are NSEC3 signed. And this patch improves that.

PR #390, opened 2024-10-18
Make sure other tasks are processed with tasklists that contain both bad xfrs (that cause reload to exit) and other tasks

wtoorop:
The issue is that updating the database from transfers and other tasks that change the database (such as adding zones or changing patterns) are processed in so called task list. Especially applying transfers are executed batch-wise periodically as configured by xfrd-reload-timeout: (default every second). If a zone is added it is appended to the tasklist. If there would have been an "apply transfer task" before that other (adding a zone) task, then that will be executed first. If the "apply transfer task" fails, the reload will be aborted (because the database could be corrupt) and the effect of any tasks before or after the transfer are undone. This can lead to a situation that the transfer daemon (xfrd) thinks a zone is configured, but the zone is not configured in the server process (main).

This PR replaces PR #332 . It removes the additional fork by letting serve childs send NOTIFY messages directly to the xfrd process instead of letting main or backup-main proxy for it.


tpkg/bad_xfr_with_other_tasks.tpkg/bad_xfr_with_other_tasks.test
```diff
@@ -0,0 +1,39 @@
+# #-- bad_xfr_with_other_tasks.test --#
+# source the master var file when it's there
+[ -f ../.tpkg.var.master ] && source ../.tpkg.var.master
+# use .tpkg.var.test for in test variable passing
+[ -f .tpkg.var.test ] && source .tpkg.var.test
+
+. ../common.sh
+PRE="../.."
+TPKG_NSD_CONTROL="$PRE/nsd-control"
+PRIMARY_CONTROL="$PRE/nsd-control -c primary.conf"
```
wcawijngaards: The primary is ldns-testns, and this nsd-control variable is not used.

```suggestion
```

wcawijngaards (approved):
Different notify communication could be an improvement, if it is easier. The task separation hopefully stops state mismatch. And the code looks nice.

wtoorop:
Just for the record. A write to a pipe is guaranteed to be atomic when the written size is less then `PIPE_BUF` ([according to POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/functions/write.html#tag_17_699_08)). The size of `PIPE_BUF` is at least 512 ([according to POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/limits.h.html)).

An alternative (simpler?) solution could be to use a single pipe to the `xfrd` where all the serve children are writing on the other end of that single pipe, and then communicate only the name of the zone and an optional serial, instead of the whole NOTIFY message, which is together less than 512.

I think that with such a solution the necessity for a "NOTIFY per zone" buffering mechanism becomes a bit more urgent though, since the serve children loose their individual pipe buffer.

k0ekk0ek (approved):
Thanks for cleaning up the forward mode :+1:

PR #395, opened 2024-10-24
Explain the zonefile example better

melroy89:
Most people who read the example zone file for the fist time are new (like me).

I believe adding these three notes would help a lot to get people started. And remove any confusion.


wcawijngaards (approved):
The additional notes are correct. But the NSD documentation is not really a guide for how to use DNS. So I can put this in, but realistically this is not an introductory text to explain how that DNS works.

melroy89:
> The additional notes are correct. But the NSD documentation is not really a guide for how to use DNS. So I can put this in, but realistically this is not an introductory text to explain how that DNS works.

Understood, but especially the missing `IN` was very confusing for me. In fact, I thought the example was wrong at first. Until I read about the implicit record type behavior.  Thanks for merging any ways!

PR #402, opened 2024-10-28
Improve build from source section

melroy89:
- Add `Sphinx-Substitution-Extensions` to pypi requirements: https://pypi.org/project/sphinx-substitution-extensions/
- Use Sphinix extension for code blocks: `sphinx_substitution_extensions`
- Rename Ubuntu 20.04 LTS to just Ubuntu
- Use substitution for the `version` number in both the text as well as the code block

---

Should we also try to upgrade [Sphinx](https://pypi.org/project/Sphinx/) to v8.1.3 next?

### DIFF
--- a/doc/manual/conf.py.in
+++ b/doc/manual/conf.py.in
@@ -63,7 +63,7 @@ extensions = [
     'sphinx.ext.autosectionlabel',
     'notfound.extension',
     'sphinxcontrib.jquery',
-    'sphinx_rtd_theme',    
+    'sphinx_rtd_theme',
     'sphinx_substitution_extensions',
 ]
 
--- a/doc/manual/conf.py.in
+++ b/doc/manual/conf.py.in
@@ -63,7 +63,8 @@ extensions = [
     'sphinx.ext.autosectionlabel',
     'notfound.extension',
     'sphinxcontrib.jquery',
-    'sphinx_rtd_theme',
+    'sphinx_rtd_theme',    
+    'sphinx_substitution_extensions',
 ]
 
 intersphinx_mapping = {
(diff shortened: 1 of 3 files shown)


k0ekk0ek:
Thanks @melroy89!

melroy89:
Oei I notice something doesn't work. Somehow with the substitution of the version number in the code block.

@k0ekk0ek But I did do it correcty right? Is the pip package installed om the server correctly? What did I miss?
Screenshot_20241028-153516

k0ekk0ek:
Oh, now I recall... I made these types of changes myself and then found out `|version|` is not supported in code blocks.

k0ekk0ek:
@melroy89, do you want me to issue a PR or do you prefer to send a follow-up PR to fix it (either is fine with me).

melroy89:
> Oh, now I recall... I made these types of changes myself and then found out `|version|` is not supported in code blocks.

Yes it is supported with this python extension. That is the whole point of introducing https://pypi.org/project/sphinx-substitution-extensions/

k0ekk0ek:
Well, the extension page says Sphinx `7.2.0+`. As we've pinned on `5.0.2`, that is presumably the problem(?)

melroy89:
> Well, the extension page says Sphinx `7.2.0+`. As we've pinned on `5.0.2`, that is presumably the problem(?)

Can we upgrade Sphinx for NSD?

melroy89:
aha you did! https://github.com/NLnetLabs/nsd/commit/89aeb685ffe8a4c73e7023f35e98b99a3b448c51 let's hope it works!?

PR #406, opened 2024-11-08
ohttp and tls-supported-groups SvcParam support

wtoorop:
Currently includes simdzone from the `svcb-tls-supported-groups` branch. I'll commit chaning that to `main` once that has been approved and merged.

### DIFF
--- a/rdata.c
+++ b/rdata.c
@@ -69,7 +69,8 @@ lookup_table_type dns_algorithms[] = {
 
 const char *svcparamkey_strs[] = {
 		"mandatory", "alpn", "no-default-alpn", "port",
-		"ipv4hint", "ech", "ipv6hint", "dohpath"
+		"ipv4hint", "ech", "ipv6hint", "dohpath", "ohttp",
+		"tls-supported-groups"
 	};
 
 typedef int (*rdata_to_string_type)(buffer_type *output,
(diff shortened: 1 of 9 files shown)


rdata.c
```diff
@@ -805,6 +806,21 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
 	return 1;
 }
 
+static int
+rdata_svcparam_tls_supported_groups_to_string(buffer_type *output,
+		uint16_t val_len, uint16_t *data)
+{
+	assert(val_len > 0); /* Guaranteed by rdata_svcparam_to_string */
+
+	if ((val_len % sizeof(uint16_t)) == 0)
```
wcawijngaards: This seems to fail multiples of uint16_t, perhaps it should check for not a multiple of uint16_t.

```suggestion
	if ((val_len % sizeof(uint16_t)) == 1)
```

wcawijngaards (approved):
That looks fine. It adds svcparam types that are supported.

rdata.c
wtoorop: Good catch! Thanks!

wtoorop: It wasn't picked up by the `svcb.tdir` test because both the presentation format and the wire format zone files translated to RFC3597 unknown type format. The two translations were then compared.
I have now manually checked that it now does print in presentation format.

k0ekk0ek (approved):
Changes look good.

PR #407, opened 2024-11-11
Tweaking logging verbosity levels

wtoorop:
So we not only see when a transfer ends (level 1), but also when it started (at verbosity level 3) And we do not only see when a zone write started (level 1), but also when it ends (with verbosity level 3).

Furthermore, you now need at least verbosity 4 to see confirmation of received XFR packets, with level 4 logging every 10000th packet, level 5 logging every 1000th packet and only level 6 logging every XFR packet (as level 3 did before)


k0ekk0ek (approved):
These changes make sense to me. Should allow for more convenient debugging.

PR #408, opened 2024-11-15
NINFO, RKEY, RESINFO, WALLET, CLA and TA RR types

wtoorop:


wcawijngaards (approved):
Nice to be able to handle additional RR types. The code looks nice, and the type TA with high number seems to be handled well, like the DLV type.

PR #409, opened 2024-11-18
Printing of NSAP-PTR, GPOS and HIP

wtoorop:
NSAP-PTR, GPOS and HIP had parsing implementation in simdzone, but could not yet be printed. With these changes they can be printed.

Furthermore, the text elements of RESINFO should be printed without quotes and a simplification of the SPF definition to match that of TXT.


k0ekk0ek (approved):

k0ekk0ek:
Good stuff @wtoorop :+1:

PR #413, opened 2024-12-13
set persist-credentials: false per zizmor suggestion

maertsen:
I don't think we currently have anything in CI that needs the (readonly) checkout credentials persisted to .git/config during the run.


wtoorop (approved):
Ack, so please go ahead and merge

PR #414, opened 2024-12-19
XoT interoperability

wtoorop:
- BIND needs the "dot" alpn (as is also required by RFC 9103)
- Knot closes the TLS session causing errors to be reported by NSD


wcawijngaards (approved):
The changes look good. Nice that it can be more interoperable.

PR #418, opened 2025-01-03
Some more additional RR types

wtoorop:
Added support for DSYNC, EID, NIMLOC, SINK, TALINK, DOA, AMTRELAY and IPN RR types.


wcawijngaards (approved):
The companion change for the simdzone additional RR types support. Nice to have the additional RR types support. They are all generic types, so this only provides zonefile syntax read and print. I am a bit uncomfortable with the reuse of the IPSECKEY gateway function for the new type AMTRELAY, because the 0x7f mask for the gateway type is not part of the IPSECKEY format.

wtoorop:
> The companion change for the simdzone additional RR types support. Nice to have the additional RR types support. They are all generic types, so this only provides zonefile syntax read and print. I am a bit uncomfortable with the reuse of the IPSECKEY gateway function for the new type AMTRELAY, because the 0x7f mask for the gateway type is not part of the IPSECKEY format.

Thanks, I agree. Last commit (4c456af) fixes just that.

Issue #419, opened 2025-01-06
"state" should be something else than "ok" then served-serial and commit-serial are not equal 


pettai:
It's a NSD behavior follow-up on https://github.com/NLnetLabs/nsd/issues/417 

`nsd-control zonestatus` reports zone `state: ok` even thought served-serial is several updates away from current commit-serial.

Example [https://github.com/NLnetLabs/nsd/issues/417](https://github.com/NLnetLabs/nsd/issues/417#issuecomment-2567196658)

Generally, perhaps it shouldn't be regarded as a fault state (since the new zone data is suppose to be installed, but currently hasn't been that yet). 
So perhaps something like `state: old-serial` or similar would be an indication that the zone currently isn't on par with the latest zone NSD knows about.

This would make it much simpler to discover then the NSD updating process isn't working as expected (as per https://github.com/NLnetLabs/nsd/issues/417)

wtoorop:
Good idea. I'll implement that shortly.

pettai:
your fix worked for us 👍

pettai:
Apparently, this can also happen for catzones too:

```
root@sunic:~# nsd-control zonestatus catz.catalog.
zone:	catz.catalog.
	catalog: consumer
	state: ok
	served-serial: "1734425182 since 2025-01-13T08:45:02"
	commit-serial: none
	wait: "80155 sec between attempts"
```

This happened after a restart of `nsd` due to the previous bug.
And none of the catz-zones where served at this state either. 

After a manual `nsd-control transfer catz.catalog.` the catz-zones started to be served.

wtoorop:
Ack, served-serial: none and commit-serial: something is okay, but not the other way around. But there are some retries implemented already. I'll have to consider this...

pettai:
If I was unclear, the issue was not just `commit-serial: none`, but the fact that NSD didn't serve any of the catzones of the catalog zone. First after I did `nsd-control transfer catz.catalog.` (no serial update) all the catalog zones got loaded and served immediately. 
I'm guessing that `catz.catalog.` was loaded from disk upon the restart of NSD, but none of the zones from the catalog zone where loaded. (Note the comment after `catalog: consumer` that wasn't present after the NSD restart)

```
zone:	catz.catalog.
	catalog: consumer (serial: 1734425182, # members: 414)
	state: ok
	served-serial: "1734425182 since 2025-01-13T19:45:01"
	commit-serial: "1734425182 since 2025-01-13T19:45:01"
[...]

PR #420, opened 2025-01-07
State "old-serial" for nsd-control

wtoorop:
As requested in Issue #419
"old-serial" is printed as state with nsd-control when the served serial is older than the one received by transfer. Also, state "future-serial" is printed if the served serial is newer than the one received by transfer.

### DIFF
--- a/remote.c
+++ b/remote.c
@@ -1123,8 +1123,13 @@ print_zonestatus(RES* ssl, xfrd_state_type* xfrd, struct zone_options* zo)
 		return 1;
 	}
 	if(!ssl_printf(ssl, "	state: %s\n",
-		(xz->state == xfrd_zone_ok)?"ok":(
-		(xz->state == xfrd_zone_expired)?"expired":"refreshing")))
+	     xz->state == xfrd_zone_expired                  ? "expired"
+	   : xz->state != xfrd_zone_ok                       ? "refreshing"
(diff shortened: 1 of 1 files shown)


pettai:
It worked for us 👍 , I've added this patch to our own build

wtoorop:
> It worked for us 👍 , I've added this patch to our own build

Ah! That's good to know! I guess it's ready for review from Wouter and Jannik then :)

wcawijngaards (approved):
Maybe this output helps debugging. The code looks good to have.

pettai:
Not sure if the `future-serial` state is possible, though.
Seems NSD ignores notifies or transfers of older serials prematurely 

```
Jan  9 10:05:34 sunic nsd[2978647]: xfrd: zone example.net ignoring old serial (2025010900/2024123101) from 130.242.125.XXX
```

And the state becomes `refreshing`. I've also tried to update serial on disk to a higher and lower value, but that doesn't seem to have any effect either. Perhaps a huge zone is needed for this to be noticed...

pettai:
Apparently, the catz zone can also end up in a non-ok state, yet reported as `state: ok` (ref https://github.com/NLnetLabs/nsd/issues/419#issuecomment-2586582571)
Perhaps that could be fixed too before this pull request gets merged?

PR #421, opened 2025-01-07
old-main can quit before informing reload

wtoorop:
When during a reload, old-main has finished all it's tasks (killing the old server processes, informing xfrd that it's going to exit etc.), it will shutdown (cmd == NSD_QUIT), but before it shuts down, it informs the reload process that it is done by sending it NSD_RELOAD over the reload_listerener. However, on some platforms, the reload_listener pipe may be closed (by old-main's) shutdown, before the reload process can read from it. This was nog logged as an error before (NSD 4.8) and neither should it now.

Thanks Otto Retter for reporting this on the nsd-users list.

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,3 +1,8 @@
+7 January 2025: Willem
+	- Fix #421: old-main can quit before the reload process received        
+	  from old-main that it is done on the reload_listener pipe.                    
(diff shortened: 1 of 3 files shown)


wcawijngaards (approved):
That order of socket activations can happen, so it makes sense to handle it as a non-failure. It is nice to have this code.

PR #422, opened 2025-01-15
draft-homburg-deleg-incremental-deleg support

WP-Official:
Enable with `--enable-deleg` to configure


query.c
```diff
@@ -949,6 +952,17 @@ answer_delegation(query_type *query, answer_type *answer)
...
 	}
+#ifdef USE_DELEG
+	if ((rrset = domain_find_deleg_rrsets(query->delegation_domain, query->zone, nsd->db, &rrset_deleg_signatures)))
+	{
+
+		add_rrset(query, answer, AUTHORITY_SECTION,
+			query->delegation_domain, rrset);
+	if (rrset_deleg_signatures && query->edns.dnssec_ok && zone_is_secure(query->zone))
+		add_rrset(query, answer, AUTHORITY_SECTION,
+			query->delegation_domain, rrset_deleg_signatures);
```
wtoorop: Hi @WP-Official , the RRs need to be added with their actual owner name. So not query->delegation_domain, but the one you composed in domain_find_deleg_rrsets.

WP-Official: This has been changed, it now uses the constructed ideleg domain as an owner for the signatures and IDELEG RRset as well!

Issue #417, opened 2024-12-27
bug: nsd stops forking with new zone data


pettai:
It seems our NSD secondary has triggered some sort of intermittent bug
This happens several weeks/months of running nsd stops forking with the new zone data.

A manual `nsd-control transfer some.zone`. or even `nsd-control force_transfer some.zone` won’t work.
Only restart of nsd solves the problem.

The logs doesn’t give a hint other than process `nsd[3185521]` stopped reporting updates.

```
[...]
Dec 27 05:41:04 sunic nsd[1413]: zone some.zone. serial 1735277408 is updated to 1735278004
Dec 27 05:46:55 sunic nsd[1413]: xfrd: zone some.zone. committed "received update to serial 1735278005 at 2024-12-27T05:46:55 from 2001:6b0:: TSIG verified with key tsig-sunet"
Dec 27 05:46:58 sunic nsd[3185521]: zone some.zone. received update to serial 1735278005 at 2024-12-27T05:46:55 from 2001:6b0:: TSIG verified with key tsig-sunet of 1369 bytes in 9.7e-05 seconds
Dec 27 05:47:02 sunic nsd[1413]: zone some.zone. serial 1735278004 is updated to 1735278005
Dec 27 05:07:59 sunic nsd[1413]: zone some.zone. serial 1735275607 is updated to 1735275608
Dec 27 05:10:42 sunic nsd[1413]: zone some.zone. serial 1735275608 is updated to 1735275609
Dec 27 05:11:04 sunic nsd[1413]: zone some.zone. serial 1735275609 is updated to 1735276204
[...]
```

And `nsd-control zone status` says `state: ok`, (which seems odd since several updates has failed to update the zone.)

```
zone:	some.zone
	state: ok
	served-serial: "1735279204 since 2024-12-27T06:01:01"
	commit-serial: "1735311605 since 2024-12-27T15:01:10"
	wait: "3510 sec between attempts"

```

Unfortunately, we restarted nsd since nothing else worked.
So we don't know if `nsd[3185521]` died or got in a defunct state.

Currently we run nsd 4.11.0 based on debians build/package , but this issue has happened with nsd 4.10.x too. 

```
# nsd -v
NSD version 4.11.0
Written by NLnet Labs.

[...]

PR #423, opened 2025-01-16
Fix reload 100% CPU spinning after failed reload

wtoorop:
After a failed reload, some UDB pointers were cleared that were still linked in the UDB. This fix unlinks them in all the correct places.

This may be the fix to issue #417

### DIFF
--- a/server.c
+++ b/server.c
@@ -2657,7 +2657,7 @@ server_reload(struct nsd *nsd, region_type* server_region, netio_type* netio,
 		exit(1);
 	}
 	assert(cmd == NSD_RELOAD);
-	udb_ptr_unlink(last_task, nsd->task[nsd->mytask]);
+	udb_ptr_set(last_task, nsd->task[nsd->mytask], 0);
 	task_process_sync(nsd->task[nsd->mytask]);
 #ifdef USE_ZONE_STATS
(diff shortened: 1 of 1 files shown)


wcawijngaards (approved):
It makes sense that on failure, the allocated objects need to be taken out of use. Hopefully that fixes the task file corruption issue.

wtoorop:
> It makes sense that on failure, the allocated objects need to be taken out of use. Hopefully that fixes the task file corruption issue.

@wcawijngaards I want to do it differently anyway.

Although I do think this will fix the 100% spinning CPU at the load process (also because the last_task is not used by the forked load process if one of the process xfr tasks fails), I do still feel uncomfortable that the udb_ptr's are implicitly copied by the fork. I would rather have them unlinked after the first iteration of processing the non-transfer tasks, so that the failure cases don't need to to touch them at all.

I'd like to try another iteration where both last_task and xfrs2process are conveyed to the load process after fork on the task list itself. Hopefully I can work on that this afternoon and we can do an emergency nsd-4.11.1 quickfix release early next week (when results from @pattai confirm this fixed the issue).

wtoorop:
Closing this in favor of the simpler approach in #424

PR #424, opened 2025-01-17
Unlink tasklist udb_ptr's after failed fork() or failed apply_xfr tasks

wtoorop:
This needs to be fixed, because otherwise the udb_ptr's are inconsistent. I have good faith that this may well be the issue causing the 100% CPU in the reload process from issue #417


wcawijngaards (approved):
Looks like it cleans up the udb ptrs. Otherwise those references are used when the udb clears up space. The code looks nice.

PR #428, opened 2025-02-07
Do not listen on plain UDP on the same port as DoT

gthess:
"Similar" patch as with Unbound to disable plain channels next to encrypted ones on the same port number.

It passes around the tls_port as to not mess much with the nsd_bitset.
Maybe a more radical approach would make sense but I am not comfortable with how the nsd_bitset is set and used.

### DIFF
--- a/nsd.c
+++ b/nsd.c
@@ -263,7 +263,8 @@ setup_socket(
 
 static void
 figure_socket_servers(
-	struct nsd_socket *sock, struct ip_address_option *ip)
+	struct nsd_socket *sock, struct ip_address_option *ip,
+	const char* tls_port)
 {
 	int i;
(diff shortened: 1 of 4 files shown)


wcawijngaards (approved):
The code looks good as it is. More fine grained control over sockets and protocols is also possible, but much larger as a change, and this fixes the issue with UDP on DNS over TLS ports.

wtoorop (approved):
Can you add a line in the man page with tls-port? Thanks!

gthess:
Thanks for the reviews!

gthess:
This PR was reverted because of unseen failures at the time; namely software that probes for the SOA record on UDP.

PR #436, opened 2025-03-20
Fix print of RR type NSAP-PTR

wcawijngaards:
Fix print of RR type NSAP-PTR, it is an unquoted string. This fix goes with the fix in simdzone, and is to be applied together with that.

The RR type NSAP-PTR is deprecated.


mozzieongit (approved):

wtoorop (approved):

wcawijngaards:
Thank you for the reviews!

PR #438, opened 2025-04-09
Documentation: add RFC compliance document.

gthess:
This is the table of the website ported over here as a single source of truth.


gthess:
Preview: https://nsd.docs.nlnetlabs.nl/en/doc-rfc-compliance/reference/rfc-compliance.html

PR #439, opened 2025-04-17
Enable compilation of prometheus metrics by default

mozzieongit:


wtoorop (approved):
Looks good to me! Don't forget to merge this into the release branch too ;)

mozzieongit:
@wcawijngaards noted that it would break when using libev instead of libevent, so I added a check for the http functionality

wcawijngaards (approved):
The code looks good. Easier to have less configure options, but because the library is maybe not there, it remains ifdef conditional code.

PR #444, opened 2025-06-03
Direct rdata storage

wcawijngaards:
Minimize rdata storage in NSD. This branch starts from the work that Jeroen did on rdata minimization. The code from the minimize-rdata-again branch is continued in development. That code can be found here
https://github.com/k0ekk0ek/nsd/tree/minimize-rdata-again

This reduces the memory used, for the zone as it is in memory for NSD. The amount of reduction depends on the zone contents. It allocates the rdata parts of Resource Records together and this reduces the memory use of rdata storage in NSD by removing
the overhead of storing the rdata in small parts.

It can be combined with `--enable-packed`, and `--disable-radix-tree`, that also reduce memory use.

Since references to domain names are stored as pointers to the name tree, just like before, the speed of service is not that different.


dname.c
```diff
@@ -620,3 +620,107 @@ is_dname_subdomain_of_case(const uint8_t* d, unsigned int len,
...
+	if(l > MAXDOMAINLEN)
+		return 0;
+	return l;
+}
+
+int
+dname_make_buffered(struct dname_buffer* dname, uint8_t *name, int normalize)
+{
+	size_t name_size = 0;
+	uint8_t label_offsets[MAXDOMAINLEN];
```
mozzieongit: Could this be `MAXDOMAINLEN/2+1`?

```suggestion
	uint8_t label_offsets[MAXDOMAINLEN/2+1];
```

dname.c
```diff
@@ -620,3 +620,107 @@ is_dname_subdomain_of_case(const uint8_t* d, unsigned int len,
...
+		if (label_is_root(label))
+			break;
+		if (name_size > MAXDOMAINLEN)
+			return 0;
+
+		label = label_next(label);
+	}
+
+	if (name_size > MAXDOMAINLEN)
+		return 0;
```
mozzieongit: You could avoid the extra check after the loop by checking the length before breaking out of the loop.
```suggestion
		if (name_size > MAXDOMAINLEN)
			return 0;
		if (label_is_root(label))
			break;

		label = label_next(label);
	}
```

dname.c
```diff
@@ -620,3 +620,107 @@ is_dname_subdomain_of_case(const uint8_t* d, unsigned int len,
...
+	buffer_type *packet, int allow_pointers, int normalize)
+{
+	int wirelen = dname_make_wire_from_packet(dname->storage,
+		packet, allow_pointers);
+	if(wirelen == 0)
+		return 0;
+	if(!dname_make_buffered(dname, dname->storage, normalize))
+		return 0;
+	/* This could be an assertion. */
+	if(wirelen != dname->dname.name_size)
```
mozzieongit: Do you want to make it an assertion?

dns.h
```diff
@@ -115,7 +120,7 @@ typedef enum nsd_rc nsd_rc_type;
 #define TYPE_EID	31	/* draft-ietf-nimrod-dns-01 */
 #define TYPE_NIMLOC	32	/* draft-ietf-nimrod-dns-01 */
 #define TYPE_SRV	33	/* SRV record RFC2782 */
-#define TYPE_ATMA	34	/* ATM Address */
+#define TYPE_ATMA	34	/* ATM Address http://www.jhsoft.com/rfc/af-saa-0069.000.rtf */
```
mozzieongit: I get a 404 on that link (and also on the link on the IANA page for that RR type)

dns.h
```diff
@@ -208,102 +211,203 @@ typedef enum nsd_rc nsd_rc_type;
...
+	 * is shorter. Also no following rdatas after it. Non optional
+	 * rdatas must be present. */
+	int is_optional;
+
+	/* The length, in bytes, of the rdata field. Can be set to
+	 * a specialized value, like RDATA_COMPRESSED_DNAME,
+	 * RDATA_STRING, ...
+	 * If there is a length function, that is used. That is for any type
+	 * where the length depends on a value in the rdata itself. */
+	int32_t length;
```
mozzieongit, line 290:
```suggestion
	/* The length, in bytes, of the rdata field. Can be set to
	 * a specialized value, like RDATA_COMPRESSED_DNAME,
	 * RDATA_STRING, ..., if there is a length function, that is used.
	 * That is for any type where the length depends on a value in
	 * the rdata itself. */
	int32_t length;
```

dns.h
```diff
@@ -208,102 +211,203 @@ typedef enum nsd_rc nsd_rc_type;
...
+	int has_references;
+	/* Whether RDATA contains compressible names. The output can be
+	 * compressed on the packet. If true, also has_references is true,
+	 * for the packet encode routine, that uses the references. */
+	int is_compressible;
+	/* The type has domain names, like literal dnames in the format. */
+	int has_dnames;
+	/* Read function that copies rdata for this type to struct rr. */
+	nsd_read_rdata_type read_rdata;
+	/* Write function, that copie rdata from struct rr to packet. */
```
mozzieongit:
```suggestion
	/* Write function, that copies rdata from struct rr to packet. */
```

query.c
```diff
@@ -1783,8 +1800,8 @@ query_add_optional(query_type *q, nsd_type *nsd, uint32_t *now_p)
 		if(q->edns.zoneversion
 		&& q->zone
 		&& q->zone->soa_rrset
-		&& q->zone->soa_rrset->rrs
-		&& q->zone->soa_rrset->rrs->rdata_count >= 3)
+		&& q->zone->soa_rrset->rr_count >= 1
+		&& q->zone->soa_rrset->rrs[0]->rdlength >= 20+2*sizeof(void*))
```
mozzieongit: Why is it `20+2*sizeof(void*)`? Where do the numbers come from?

mozzieongit: Ah, it's SOA's 5*4(bytes for the numbers)+2 (pointer to names)

xfrd-catalog-zones.c
```diff
@@ -545,11 +546,11 @@ xfrd_process_catalog_consumer_zone(
...
-		if (rrset->rrs[i].rdata_count != 1)
-			continue;
-		if (rrset->rrs[i].rdatas[0].data[0] == 2
-		&&  ((uint8_t*)(rrset->rrs[i].rdatas[0].data + 1))[0] == 1
-		&&  ((uint8_t*)(rrset->rrs[i].rdatas[0].data + 1))[1] == '2') {
+		if(rrset->rrs[i]->rdlength > 1 &&
+		   rrset->rrs[i]->rdata[0] == 1 &&
+		   rrset->rrs[i]->rdlength ==
+			((uint16_t)rrset->rrs[i]->rdata[0]) + 1 &&
+		   rrset->rrs[i]->rdata[1] == '2') {
```
mozzieongit: This could be simplified:

```suggestion
		if(rrset->rrs[i]->rdlength == 2 &&
		   rrset->rrs[i]->rdata[0] == 1 &&
		   rrset->rrs[i]->rdata[1] == '2') {
```

But I'm suspicious about the `rrset->rrs[i]->rdata[0] == 1` and am not entirely sure about what it is supposed to mean.

mozzieongit: Oh, it's because it is a TXT RR that has its length at position 0, right?

tpkg/cutest/cutest_namedb.c
```diff
@@ -647,6 +657,7 @@ parse_rr_str(struct zone *zone, char *input, struct parse_rr_state *state)
 
 	length = strlen(input);
 	string = malloc(length + 1 + ZONE_BLOCK_SIZE);
+	memset(string, 0, length + 1 + ZONE_BLOCK_SIZE);
 	memcpy(string, input, length);
 	string[length] = 0;
 	memset(string+length+1, 0, ZONE_BLOCK_SIZE);
```
mozzieongit: One of the memsets could be removed as they are zeroing the same data.

tpkg/cutest/cutest_dns.c
```diff
@@ -15,27 +15,161 @@
...
-		CuAssert(tc, "dns rrtype descriptor: offset", i == d - start);
+		const struct nsd_type_descriptor* d = start+i;
+		const struct nsd_type_descriptor* lookup = nsd_type_descriptor(
+			d->type);
+
+		if(i <= 264) {
+			CuAssert(tc, "dns rrtype descriptor: index",
+				i == d->type);
+			CuAssert(tc, "dns rrtype descriptor: offset",
+				i == d - start);
```
mozzieongit: Given:
```
		const struct nsd_type_descriptor* d = start+i;
```
The following is always true, as `d` is set to `start+i` above:
```
			CuAssert(tc, "dns rrtype descriptor: offset",
				i == d - start);
```

difffile.c
```diff
@@ -323,112 +314,172 @@ rrset_delete(namedb_type* db, domain_type* domain, rrset_type* rrset)
...
-				domain_dname(b[k].domain))!=0) {
-				*rdnum = k;
+
+	if (memcmp(rr1->rdata, rr2->rdata, rr1->rdlength) == 0)
+		return 1;
+
+	descriptor = nsd_type_descriptor(type);
+	for (size_t i=0; i < descriptor->rdata.length && offset < rr1->rdlength; i++) {
+		uint16_t field_len1, field_len2;
+		struct domain* domain1, *domain2;
```
mozzieongit:
```suggestion
		struct domain *domain1, *domain2;
```

mozzieongit (commented):
I still have to read the `rdata.[ch]` files, but I'll submit the current notes already in case you want to have a look already.

dname.c
wcawijngaards: Yes it looks like it should be. It was copied from dname_make() earlier, that had the original max.

wcawijngaards: In https://github.com/NLnetLabs/nsd/pull/444/commits/48fde5528f09409ee3228264840a836e6eaff6ee it also fixes dname_make in the same way.

dname.c
wcawijngaards: Yes that could be a little faster too! I have also fixed dname_make() in the same way, in https://github.com/NLnetLabs/nsd/pull/444/commits/6543784f49b583e4aebcfb1e81c670cfcdeccd8e .

dname.c
wcawijngaards: The commit added the assertion.

dns.h
wcawijngaards: Fixed in the commit.

dns.h
wcawijngaards: Fixed with commit.

dns.h
wcawijngaards: Fixed spelling with commit, thank you!

query.c
wcawijngaards: Added a comment to the line, and one below.

xfrd-catalog-zones.c
wcawijngaards: Yes it is the TXT string length, of the first string rdata element. Added a comment for it, in the commit.

tpkg/cutest/cutest_namedb.c
wcawijngaards: Removed it, in the commit.

tpkg/cutest/cutest_dns.c
wcawijngaards: Removed the unused assertion from the unit test, in the commit.

rdata.c
```diff
@@ -67,414 +69,455 @@ lookup_table_type dns_algorithms[] = {
...
-rdata_dname_to_string(buffer_type *output, rdata_atom_type rdata,
-	rr_type* ATTR_UNUSED(rr))
+print_name_literal(struct buffer *output, uint16_t rdlength,
+	const uint8_t *rdata, uint16_t *offset)
 {
-	buffer_printf(output,
-		      "%s",
-		      dname_to_string(domain_dname(rdata_atom_domain(rdata)),
-				      NULL));
+	const uint8_t* name, *label, *limit;
```
mozzieongit:
```suggestion
	const uint8_t *name, *label, *limit;
```

rdata.c
```diff
@@ -67,414 +69,455 @@ lookup_table_type dns_algorithms[] = {
...
-	}
-	return 1;
-}
-
-static int
-rdata_long_text_to_string(buffer_type *output, rdata_atom_type rdata,
-	rr_type* ATTR_UNUSED(rr))
+static int32_t
+print_text(struct buffer *output, uint16_t rdlength, const uint8_t *rdata,
+	uint16_t *offset)
```
mozzieongit, line 262: Is there a reason the `print_text` function doesn't do the "offset is less than rdlength check" and early return like the other printing functions above?
`print_string` prints nothing (and return 0) if offset is too large, while `print_text` prints `""` (and always returns 1)

rdata.c
```diff
@@ -67,414 +69,455 @@ lookup_table_type dns_algorithms[] = {
...
 				buffer_printf(output, "\\");
 			}
 			buffer_printf(output, "%c", ch);
 		} else {
-			buffer_printf(output, "\\%03u", (unsigned) data[i]);
+			buffer_printf(output, "\\%03u", (unsigned) rdata[i]);
 		}
 	}
 	buffer_printf(output, "\"");
+	*offset += (rdlength - *offset);
```
mozzieongit:
```suggestion
	*offset = rdlength;
```

rdata.c
```diff
@@ -67,414 +69,455 @@ lookup_table_type dns_algorithms[] = {
...
-
-static int
-rdata_long_to_string(buffer_type *output, rdata_atom_type rdata,
-	rr_type* ATTR_UNUSED(rr))
+print_ip4(struct buffer *output, size_t rdlength, const uint8_t *rdata,
+	uint16_t *offset)
 {
-	uint32_t data = read_uint32(rdata_atom_data(rdata));
-	buffer_printf(output, "%lu", (unsigned long) data);
+	char str[INET_ADDRSTRLEN + 1];
```
mozzieongit, line 338: Do you want to add the assertion here too, like in the other print functions below?
```suggestion
	char str[INET_ADDRSTRLEN + 1];
	assert(rdlength >= *offset);
```

rdata.c
```diff
@@ -67,414 +69,455 @@ lookup_table_type dns_algorithms[] = {
...
+
 	buffer_reserve(output, size * 2 + 1);
-	length = b32_ntop(rdata_atom_data(rdata)+1, size,
-			  (char *) buffer_current(output), size * 2);
-	if (length > 0) {
-		buffer_skip(output, length);
-	}
-	return length != -1;
+	length = b32_ntop(rdata + *offset + 1, size,
+			  (char *)buffer_current(output), size * 2);
```
mozzieongit, line 486: Why is it `b32_ntop(rdata ... (output), ` with `size * 2)` instead of `size * 2 + 1` as reserved above?

mozzieongit: Same question for `print_base64`

rdata.c
```diff
@@ -683,152 +663,137 @@ rdata_nsec_to_string(buffer_type *output, rdata_atom_type rdata,
...
-	}
-
-	return length != -1;
-}
-
-static int
-rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
-	uint16_t *data)
+print_svcparam_alpn(struct buffer *output, uint16_t svcparamkey,
+	const uint8_t* data, uint16_t datalen)
```
mozzieongit, line 785: `print_svcparam_dohpath` and `print_svcparam_alpn` have different escaping for backslash and quotes. I did a quick test with writing out zonefile-updates as ixfr files and got the following:

*SVCB RR from input zonefile:*
```
sv SVCB 3 . alpn="\065\\bla,\"hi" dohpath="/\065\\bla,\"hi{?dns}"
```

*RR from ixfr file:*
```
sv.example.com. 240 IN  SVCB    3 . alpn="Abla,\\\"hi" dohpath="/A\\bla,\"hi{?dns}"
```
Here you can see that the `\\` in the alpn param disappears, while the `\\` in dohpath remains. Additionally, alpna turns `\"` into `\\\"`.

*RR from dig output:*
```
sv.example.com.     240 IN  SVCB    3 . alpn="Abla,\"hi" key7="/A\\bla,\"hi{?dns}"
```
For the dig output the alpn escape removing `\\` also seems wrong.

While having such escapes in strings is fairly stupid, as far as I can tell, they seem to be allowed.

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+			return 0;
+		*field_len = l;
+	} else if(field->length >= 0) {
+		*field_len = field->length;
+		*domain = NULL;
+	} else {
+		size_t dlen;
+		int32_t flen;
+		if(offset > rr->rdlength)
+			return 0;
```
mozzieongit, line 1072: Is there a reason to have this check only on this else branch and not for all branches above the if?
(same for `lookup_rdata_field_entry_uncompressed_wire`)

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+		if(offset > rdlength)
+			return 0;
+		/* Handle the specialized value cases. */
+		switch(field->length) {
+		case RDATA_COMPRESSED_DNAME:
+		case RDATA_UNCOMPRESSED_DNAME:
+			/* It is not a reference, but in uncompressed
+			 * wireformat in the rdata buffer, for
+			 * RDATA_COMPRESSED_DNAME and
+			 * RDATA_UNCOMPRESSED_DNAME. */
```
mozzieongit: I don't understand this comment. Does it mean these two cases cannot happen?

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+	if (!(*rr = region_alloc(domains->region, size)))
+		return TRUNCATED;
+	primary_domain = domain_table_insert(domains, (void*)&primary);
+	primary_domain->usage++;
+	mailbox_domain = domain_table_insert(domains, (void*)&mailbox);
+	mailbox_domain->usage++;
+
+	memcpy((*rr)->rdata, &primary_domain, sizeof(void*));
+	memcpy((*rr)->rdata + sizeof(void*), &mailbox_domain, sizeof(void*));
+	buffer_read(packet, (*rr)->rdata + 2 * sizeof(void*), 20);
```
mozzieongit: While that is unlikely to ever be an issue, do you want to keep it consistent with the other sizeof pointers calculations?

```suggestion
	size = sizeof(**rr) + 2 * sizeof(void*) + 20;
	if (!(*rr = region_alloc(domains->region, size)))
		return TRUNCATED;
	primary_domain = domain_table_insert(domains, (void*)&primary);
	primary_domain->usage++;
	mailbox_domain = domain_table_insert(domains, (void*)&mailbox);
	mailbox_domain->usage++;

	memcpy((*rr)->rdata, &primary_domain, sizeof(void*));
	memcpy((*rr)->rdata + sizeof(void*), &mailbox_domain, sizeof(void*));
	buffer_read(packet, (*rr)->rdata + 2 * sizeof(void*), 20);
```

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+		return MALFORMED;
+	return read_rdata(domains, rdlength, packet, rr);
+}
+
+/* Print the cm for LOC. */
+static void
+loc_cm_print(struct buffer* output, uint8_t mantissa, uint8_t exponent)
+{
+	uint8_t i;
+	/* is it 0.<two digits> ? */
```
mozzieongit, line 2120: I assume this question means, whether printing the LOC size value as "0.<calculated-value>" is correct.

The RFC uses `siz, hp, vp: [0 .. 90000000.00] (size/precision in meters)` in their zonefile syntax list, and refer to 90000km as the maximum value. However, they do not disallow a size of `0 * 10^0` (only `0 * 10^(some non-zero exponent)`), so the value 0 is allowed too.

For simpler code: Yes, it is `0.<two-digits>`. If you would want it to print something a human would write, you would check if the mantissa is zero and the exponent is zero and then only print `0` instead of `0.00`. But given that it is mathematically identical, I think we don't care.

So `0.<two-digits>` is good.

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+
+void
+write_naptr_rdata(struct query *query, const struct rr *rr)
+{
+	struct domain *domain;
+	const struct dname *dname;
+	uint16_t length;
+
+	/* short + short + string + string + string + uncompressed name */
+	assert(rr->rdlength < 7 + sizeof(void*));
```
mozzieongit:
```suggestion
	assert(rr->rdlength >= 7 + sizeof(void*));
```

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+	buffer_write(query->packet, rr->rdata, length);
+	buffer_write(query->packet, dname_name(dname), dname->name_size);
+}
+
+int
+print_naptr_rdata(struct buffer *output, const struct rr *rr)
+{
+	uint16_t length = 4;
+
+	assert(rr->rdlength > 4);
```
mozzieongit:
```suggestion
	assert(rr->rdlength >= 7 + sizeof(void*));
```

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+}
+
+void
+write_kx_rdata(struct query *query, const struct rr *rr)
+{
+	struct domain *domain;
+	const struct dname *dname;
+
+	/* short + uncompressed name */
+	assert(rr->rdlength != 2 + sizeof(void*));
```
mozzieongit:
```suggestion
	assert(rr->rdlength == 2 + sizeof(void*));
```

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+	switch (address_family) {
+	case 1: af = AF_INET; break;
+	case 2: af = AF_INET6; break;
+	}
+
+	if (af == -1 || size - 4 < length)
+		return 0;
+
+	memset(address, 0, sizeof(address));
+	memmove(address, rdata + *offset + 4, length);
```
mozzieongit, line 2611: If this were performance critical, `memcpy` would be the better choice, and safe to use as the two memory regions don't overlap, right?

No change needed, I'm more curious.

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+		return MALFORMED;
+	buffer_skip(packet, length);
+	if (!dname_make_from_packet_buffered(&target, packet,
+		1 /* lenient */, 1))
+		return MALFORMED;
+	if(rdlength < buffer_position(packet) - mark)
+		return MALFORMED;
+	length = buffer_position(packet)-mark;
+	/* For secondary, the server should accept the wireformat more
+	 * leniently. So this check is skipped. */
```
mozzieongit, line 3297: Where is it specified that the secondary should be more lenient?

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+	if(rr->rdlength < length)
+		return 0;
+	relay_type = rr->rdata[1]&AMTRELAY_TYPE_MASK;
+	buffer_printf(output, "%" PRIu8 " %c %" PRIu8,
+		rr->rdata[0],
+		(rr->rdata[1] & AMTRELAY_DISCOVERY_OPTIONAL_MASK ? '1' : '0'),
+		relay_type);
+	switch(relay_type) {
+	case AMTRELAY_NOGATEWAY:
+		break;
```
mozzieongit, line 3754: From https://www.rfc-editor.org/rfc/rfc8777.html#section-4.3.1-2:

> If the relay type field is 0, the relay field MUST be ".".

```suggestion
	case AMTRELAY_NOGATEWAY:
		buffer_printf(output, " .");
		break;
```

rdata.c
```diff
@@ -67,414 +69,455 @@ lookup_table_type dns_algorithms[] = {
...
+			/* space for labellen, label and a next root label. */
+			if (label - name > MAXDOMAINLEN-1
+				|| *label > MAXLABELLEN
+				|| limit - label < 2 + *label)
+				return 0;
+			label += 1 + *label;
+		} while (*label);
+	} else {
+		/* root domain. */
+		if(limit - label < 1)
```
wtoorop: Can this happen? That would mean the label starts beyond the rdata length, right? Is this not already covered by line 139 and 140? Maybe not if the assert isn't evaluated (in production builds), but perhaps the check can then be done earlier.

rdata.c
mozzieongit: Apparently it is also not supported by simdzone: 

log: `nsd[575856]: error: example.com.zone:18: Trailing data in AMTRELAY`
zonefile RR: `label AMTRELAY 10 0 0 .`

`label AMTRELAY 10 0 0` parses

(Note to self: offending function is `static int32_t parse_amtrelay_rdata` in `src/generic/types.h` not parsing a `.` when type is `0`)

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+	return 0;
+}
+
+int
+compare_rr_rdata(const nsd_type_descriptor_type *descriptor,
+	const struct rr *rr1, const struct rr *rr2)
+{
+	size_t index1 = 0, index2 = 0;
+	uint16_t offset1 = 0, offset2 = 0, pos = 0, field1_pos = 0,
+		field2_pos = 0, field1_len = 0, field2_len = 0;
```
mozzieongit: `field1_pos` and `field2_pos` are never changed and remain `0`

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+		return -1;
+	if(len1 > len2)
+		return 1;
+	/* len1 == len2 and equal. */
+	return 0;
+}
+
+int
+compare_rr_rdata(const nsd_type_descriptor_type *descriptor,
+	const struct rr *rr1, const struct rr *rr2)
```
mozzieongit: It looks like this whole function is never used. Is that intentional?

rdata.c
mozzieongit: I'll skip reviewing this function further until this and the unused-function question are resolved.

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+		if((rr1->rdlength == offset || rr2->rdlength == offset) &&
+			descriptor->rdata.fields[i].is_optional) {
+			/* There are no more rdata fields. */
+			/* Check lengths. */
+			if(rr1->rdlength < rr2->rdlength)
+				return 0;
+			if(rr1->rdlength > rr2->rdlength)
+				return 0;
+			/* It is equal. */
+			return 1;
```
mozzieongit:
```suggestion
			if(rr1->rdlength == rr2->rdlength)
				return 1;
			else
				return 0;
```

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+			&field_len2, &domain2))
+			malf2 = 1; /* malformed rdata buffer */
+		if(malf1 || malf2) {
+			/* Malformed entries sort last, and are sorted
+			 * equal with other malformed entries. */
+			if(!malf1 && malf2)
+				return 0;
+			if(malf1 && !malf2)
+				return 0;
+			return 1;
```
mozzieongit:
```suggestion
			if(malf1 && malf2)
				return 1;
			else
				return 0;
```

rdata.c
mozzieongit: I just notice that apart from this switch's case and the access to the rdata/rdlength (function parameters vs `rr->...`), the functions `lookup_rdata_field_entry_uncompressed_wire` and `lookup_rdata_field_entry` are pretty much identical.

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
 
+struct dname* retrieve_cname_ref_dname(const struct rr* rr)
+{
+	struct domain* domain;
+	assert(rr->type == TYPE_CNAME);
+	domain = retrieve_rdata_ref_domain(rr);
+	if(!domain)
+		return NULL;
+	return domain_dname(domain);
+}
```
mozzieongit, line 4198: Apart from the `retrieve_soa_rdata_{serial,minttl}` functions, these retrieve functions don't seem to be used anywhere. Is that intentional?

rdata.c
```diff
@@ -845,515 +810,3566 @@ rdata_svcparam_alpn_to_string(buffer_type *output, uint16_t val_len,
...
+					/* not the same length dnames. */
+					return 0;
+				}
+				if(!dname_equal_nocase(
+					(uint8_t*)dname_name(domain_dname(
+						domain1)),
+					(uint8_t*)rr2_rdata+offset2,
+					rr2_rdlen-offset2)) {
+					/* name comparison not equal. */
+					return 0;
```
mozzieongit, line 4050: Should this use the `field_len[12]` or the `buf_dname_length`-result instead of the "remainder" (`rdlen-offset`) for the dname comparisons? Same for the `dname_equal_nocase` in the `else { if(domain2)`  below?

```suggestion
				uint16_t dname_len2 = buf_dname_length(
					rr2_rdata+offset2,
					rr2_rdlen-offset2);
				if(domain_dname(domain1)->name_size !=
					dname_len2) {
					/* not the same length dnames. */
					return 0;
				}
				if(!dname_equal_nocase(
					(uint8_t*)dname_name(domain_dname(
						domain1)),
					(uint8_t*)rr2_rdata+offset2,
					dname_len2)) {
					/* name comparison not equal. */
					return 0;
```

mozzieongit (commented):
I finally finished the review of the `rdata.c` file

rdata.c
wcawijngaards: Yes, they were first made to make fast access to the contents of these record types. But then the even faster inline lookup function, from the rdata.h header was added, and these turn out to not be used so much. The compression routine uses the inline functions.

rdata.c
wcawijngaards: The variables are added because if rdata fields are of a different length, say domain names, in the two RRs, then it can keep track where the currently looked at field starts. It is a position, from 0..length of rdata. But they are not assigned I see. I also see that this function is unused, currently. Perhaps it would be fixed with an assignment, field1_pos = pos; and field2_pos = pos; in the respective if-statements, where it picks up the wireformat for the next field,  3934, 3962. That is just after, `if(field1_pos + field1_len <= pos && index1 < descriptor->rdata.length) {`. Or the function could be removed, I wrote it for having it possible that RRs can be compared.

rdata.c
```diff
@@ -67,414 +69,455 @@ lookup_table_type dns_algorithms[] = {
...
-	if (length > 0) {
-		buffer_skip(output, length);
-	}
-	return length != -1;
+	if (length == -1)
+		return 0;
+	buffer_skip(output, length);
+	*offset += size;
+	return 1;
 }
```
wtoorop, line 523:
```suggestion

/* OpenSSL had a hex_to_string function before version 1.1.0. After 1.1.0
 * the name became a define to map it to OPENSSL_buf2hexstr. The define caused
 * a function signature collision on some systems.
 */
#undef hex_to_string
```
I don't know why or what caused this, but on the system I used to evaluate memory reduction, the direct-rdata-storage branch caused a function signature collision for hex_to_string, something the master branch does not do. In any case, given that the define is present in OpenSLL after 1.1.0, I think it is safest to just undef it before defining. Alternatively we could use a different function name.

rdata.c
wcawijngaards: Sure, if you like that more, I can apply the suggestion.

rdata.c
wcawijngaards: I think due to contract change by the function, when it changed, it ended this way.

rdata.c
wcawijngaards: Certainly, the assertions can prove useful. And similarity with the other functions is good too.

rdata.c
wcawijngaards: It is fine to have the change. The consistency is nice to have.

rdata.c
wcawijngaards: That would have failed the assertion when the function is used. Good to have the fix.

rdata.c
wcawijngaards: Nice to have it similar to the other function for this type.

rdata.c
wcawijngaards: That would have failed the assertion when the function is used. Good to have the fix.

rdata.c
wcawijngaards: Applied, it seems that should be a '.'. That means the other software should also be adjusted to correctly handle the nogateway case.

rdata.c
wcawijngaards: It is shorter, but I preferred to first handle the inequality when writing the function, it is fine to have the change.

rdata.c
wcawijngaards: It is shorter, but I preferred to first handle the inequality when writing the function, it is fine to have the change.

rdata.c
wcawijngaards: I also committed the second case in the if-statement below. That is neater to use the calculated dname length.

rdata.c
wcawijngaards: Maybe because of types that have the remainder of the rdata as a text element. Like type 'URI'. So then this type of length check would fail possibly wrongly, and `""` is what is needed.

wtoorop (commented):
One question and one suggestion for something that emerged when testing this branch (function signature collision with OpenSSL for hex_to_string) that doesn't happen for the master branch.

rdata.c
wcawijngaards: The b64_ntop function, in compat/b64_ntop.c, sets the zero terminator at 'datalength' upon exit but does not count it in the returned length. I think the treatment of an ending zero terminator is what is causing this. For the buffer print, that space needs to be reserved.

rdata.c
wcawijngaards: I believe the escape and quote rules for svcb strings are unusual, so it may not be the expected escaping. It would need to be checked, but the difference for `\\\"` and `\"` with dig alpn seems to indicate that something is wrong.

wcawijngaards: The spec in the RFC says https://www.rfc-editor.org/rfc/rfc9460.html#appendix-D.2 and that indicates that the code is fine for print alpn, with `\\\<num>` for `"` and `\`, and with `\\,` for `,`.

rdata.c
mozzieongit: I see. Ok.

From the URI RFC: "The Target MUST NOT be an empty URI ("")." so it would be fine to omit the the quotes as the string is not allowed to be empty.
However, the CAA RFC does not mention whether it is disallowed to be empty, but specifies the text field to be "[...] expressed as either (1) a contiguous set of characters without interior spaces or (2) a quoted string", and as we're using a quoted string, we probably have to print the `""` even if it is empty.

So printing the empty quotes is fine and maybe even necessary.

rdata.c
wcawijngaards: It is to protect the memory references below in that code branch. I guess it could also be there as an assertion. Not that having the check in the other code would not be fine too, but the memcpy and so code is intended to be safeguarded.

rdata.c
wcawijngaards: I updated the comment, the two cases happen. The dname is then going to be in uncompressed wireformat in the uncompressed wireformat rdata buffer.

rdata.c
wcawijngaards: I also think that 0.00m printout is okay, this also indicates the precision a little. The question is referring to what the if condition tests in the line below, I think.

rdata.c
wcawijngaards: Since the offset may not be aligned, the memmove area, although not overlapping, could be unaligned to word boundaries. So I would keep it as memmove, also as print_apl is not performance critical; the APL type is not used that much also the print functions not that performance critical that would be more the DNS query handling.

rdata.c
wcawijngaards: We want the secondary configured server to be more lenient. That is because when the secondary fetches a zone from the primary, that has parsed the zone file. And has had the chance to present a parse error to the operator. Having the secondary make parse errors is not that nice, it would be better to have them printed in the primary. That is why we decided to have the secondary be more lenient, so that it will consume the zone that it accepted from zone transfer, in wireformat, once it has been printed and then possibly parsed back in, without errors.

rdata.c
wcawijngaards: I seem to have answered that in the other comment. The intent was to have code for comparison, but it was then unused. It shows how a comparison function would fit in the minimal rdata storage method.

rdata.c
wcawijngaards: Yes they are, but I am not really that keen on merging them in some way, unless it becomes more readable code. The uncompressed format needs to be dealt with by handling the same manner of cases as exist for the not uncompressed wireformat case.

rdata.c
wcawijngaards: Ok, I committed a fix to remove this check and note that it is already covered.

rdata.c
wcawijngaards: The main branch also has hex_to_string. But I renamed the hex_to_string function to buffer_print_hex in a commit, that should stop function collision.

rdata.c
mozzieongit: In its current state, I think this function would better be removed, unless you plan to fix it before merging.
When we need to compare RRs, we can add it again.

rdata.c
mozzieongit: Ok

rdata.c
wcawijngaards: I removed the unused function. Since it is not used, that code is not needed.

mozzieongit (approved):
Looks good to me.

Are you planning to fix the macOS build-test failure?

It complains about something unrelated:
```
clang --analyze -I/opt/homebrew/opt/libevent/include -I. -I/opt/homebrew/opt/openssl/include -Isimdzone/include -I./simdzone/include server.c
server.c:589:5: warning: The 1st argument to 'lseek' is -1 but should be >= 0 [unix.StdCLibraryFunctions]
  589 |         if(lseek(nsd->zonestatfd[1], (off_t)sz-1, SEEK_SET) == -1) {
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

wcawijngaards:
Thank you for the review! That found a number of issues. The build failure can be solved with merge from the main branch, but that would reset review progress, so I held off on that.

PR #445, opened 2025-06-12
contrib/nsd.openrc.in: use supervise-daemon and add `need net`

lanodan:
* supervise-daemon for restarts after a crash
* `need net` to safely bind to specific addresses/interfaces rather than ::0


wcawijngaards (approved):
This look okay! The restart after a failure is good to have. The dependency on net looks fine, if that means the interfaces can be used, and it does not look like a dependency loop.

PR #447, opened 2025-07-07
Minimize disruptions on reconfig

jaredmauch:
reconfig is called every 5 minutes from cron in my use case, the intent here is to reduce the negative impacts if the configuration as generated from database is not changed


wcawijngaards (commented):
There seems to be a bug where it uses `xfrd->need_to_send_shutdown` instead of the reload related variable? But otherwise, the changes seem fine; I would want to make a number of changes. It is nice to have the improvements!

For the simdzone change, that would need to be changed in the simdzone repository.

PR #448, opened 2025-07-07
Transfers could be more than 4GB ...

wtoorop:
and therefore need a 64bit integer to count the total size.


wcawijngaards (approved):
That helps to track the size, and not overflow on a 32bit system. It looks fine.

PR #449, opened 2025-07-14
Add what I think is useful logging

jaredmauch:
Add logging for the data I think one might want in XoT

### DIFF
--- a/xfrd-tcp.c
+++ b/xfrd-tcp.c
@@ -24,13 +24,115 @@
 #include "xfrd.h"
 #include "xfrd-disk.h"
 #include "util.h"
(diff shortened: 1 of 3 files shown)


wcawijngaards (approved):
Nice to have these additional information items printed for zone transferred over TLS. The allocation leaks into the xfrd global region, and could be in a temporary region. That would simplify things, the SSL connection is available at the printout time for information.

wcawijngaards:
In the commit, there are some fixes; configure detection and bug fixes, and it simplified by lookup of the cert information at printout time. Thank you for the addition of the TLS information for the log string for zone transfers!

PR #453, opened 2025-08-03
use pkg-config for OpenSSL

lanodan:
- configure.ac: use pkg-config for OpenSSL

Incidentally support for it got added in OpenSSL 0.9.7 which is already required.
 
https://github.com/openssl/openssl/releases/tag/OpenSSL_0_9_7
> Support for pkg-config.

- acx_nlnetlabs.m4: cleanup now unused functions for OpenSSL


wcawijngaards:
I do not really want a reliance on pkg-config. That seems to be what is there on this platform, but is not nice as a non-optional dependency. As an optional depedency, where it could be used if present, that would be acceptable. That would not remove the, then not unused, functions about ssl. Also, this is likely to break package install for most; even more so if the pkg-config entry for openssl on the system contains flags that are unexpected, I mean different from what is used to package now.

So it sure removes a bunch of code and simplifies, but the dependency on another build tool is not really that nice. It is nicer to have less dependencies for building from source.

lanodan:
Given the existing one is defined as `CHECK_SSL` maybe it could be kept as fallback in case detection via pkg-config fails?

> Also, this is likely to break package install for most; even more so if the pkg-config entry for openssl on the system contains flags that are unexpected, I mean different from what is used to package now.

Not really sure how that would happen, specially given how frequently pkg-config is used by most buildsystems out there.  
Meanwhile it would be likely to have missing flags with the existing heuristics, and almost silently so given autotools redirects compiler output into `config.log`.

PR #454, opened 2025-08-03
handle rare case but seen in production where data->query is NULL

jaredmauch:
Have seen this in production where we may see nsd server crash


wcawijngaards (approved):
The changes look good. The rbtree fix is nice to IXFR handling, and the query NULL check fix looks fine as well. A change in simdzone seems unintentional, and I likely can put that to the latest simdzone version.

PR #455, opened 2025-08-17
Add --with-dbdir to allow configuration of dbdir

jas4711:
Hi!  In Debian we used a local hack to set the dbdir for a long time:

https://sources.debian.org/src/nsd/4.12.0-1/debian/patches/0002-Force-dbdir-to-be-var-lib-nsd-by-patching-configure.patch

This pull request allows configuration of this hard coded path using `--with-dbdir` instead.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -82,6 +82,9 @@ AC_SUBST(logfile)
 # Database directory
 #
 dbdir=${localstatedir}/db/nsd
+AC_ARG_WITH([dbdir],
+	AS_HELP_STRING([--with-dbdir=dir],[NSD database directory]),
+	[dbdir=$withval])
 
 #
 # Determine the pidfile location. Check if /var/run exists, if so set pidfile


wtoorop:
Thanks @jas4711 , but nsd no longer uses an on-disk database since version 4.8.0 (release on 6 december 2023). See 526a8ce

jas4711:
But the `dbdir` variable is used throughout the `./configure.ac` file to influence other variables.  So hard-coding the path causes issues.  Or am I misunderstanding something?

wtoorop:
Oh, yes, it also defines the location for the pidfile, the xfrdstate file, the zonelistfile and the cookiesecrets file. Let me reconsider.

wtoorop:
I'll merge with slightly adapted help text.

jas4711:
Thank you for review and improvement!  Eventually Debian will pick this up too, dropping this old patch to reduce upstream diffs.

PR #457, opened 2025-08-26
Fix tests in preparation for release 4.13.0

mozzieongit:
Some fixes for tests and added a short readme for using the test scripts.


wcawijngaards (approved):
Looks okay. These types of fixes can also be committed, and that is what happened previously for it. The changes look fine, I see nothing wrong with it.

Issue #458, opened 2025-09-06
hurd-i386 compilation failure: dnstap/dnstap.c:568:34: error: implicit declaration of function ‘fstrm_tcp_writer_options_init’


jas4711:
Hi!  We built nsd 4.13.0 in Debian for many architectures, see:

https://buildd.debian.org/status/package.php?p=nsd

The only platform that is failing is hurd-i386, which I found a bit surprising since hurd-amd64 built well.  The build log is here:

https://buildd.debian.org/status/fetch.php?pkg=nsd&arch=hurd-i386&ver=4.13.0-1&stamp=1756971634&raw=0

Do you have any ideas?  Could there be some #ifdef hackery triggering a missing declaration on this particular platform only?

/Simon

wcawijngaards:
The compiler can not find the function. In this release dnstap is enabled. The libfstrm version in the compile logs is 0.3.0, but in the changelogs for libfstrm, I see that the tcp writer functions are introduced in 0.4.0, (the latest is 0.6.x), here https://github.com/farsightsec/fstrm/releases/tag/v0.4.0 .

This seems to be the failure reason. The libfstrm version is old. The solution would be to upgrade libfstrm, or with configure `--disable-dnstap` that turns off dnstap.

jas4711:
Wonderful, thanks for debugging and identifying this!  I didn't realize Debian had different libfstrm versions depending on architecture.

wcawijngaards:
Thank you for the fixup patch! With that I think I can close the issue ticket.

I modified the configure script, so that the check is performed in addition to the existing checks. So it first links with libfstrm, and then checks for additional functions that need to be available. The particular function then prints an error if not available. In this way it makes the configure script perform more extensive checks.

jas4711:
Thank for merge and further improvements!

PR #459, opened 2025-09-10
Check for libfstrm version >= 0.4 (#458)

jas4711:
This should reject too old libfstrm versions.  See #458


wcawijngaards (approved):
This looks like a good check to add. It adds a nice check for an extra function, that the code uses, from more recent libfstrm versions.

PR #462, opened 2025-09-15
Fix confusing report for default send and receive buffer-size by nsd-checkconf

mozzieongit:
Jaap reported that although the release notes mention that the default for `send-buffer-size` has been changed to 4m, he doesn't see that happen and `nsd-checkconf -o send-buffer-size ./nsd.conf.sample` reports the buffer size to be zero while the default value in the example config is set to `# send-buffer-size: 4194304`.
I assume he just missed the part in the man page that states that the `send-buffer-size` value of `0` means to use the default, but this could still be a bit clearer. 
So maybe this change is appropriate? Or should the config option simply be removed when the user wants the default value (although that would break backwards compatibility)?

### DIFF
--- a/configparser.y
+++ b/configparser.y
@@ -295,9 +295,25 @@ server_option:
   | VAR_IP_FREEBIND boolean
     { cfg_parser->opt->ip_freebind = $2; }
   | VAR_SEND_BUFFER_SIZE number
-    { cfg_parser->opt->send_buffer_size = (int)$2; }
+    {
+      if ($2 > 0) {
+        cfg_parser->opt->send_buffer_size = (int)$2;
+      } else if ($2 == 0) {
+        /* do nothing and use the default value */
+      } else {
+        yyerror("expected a number equal to or greater than zero");
+      }
+    }
   | VAR_RECEIVE_BUFFER_SIZE number
-    { cfg_parser->opt->receive_buffer_size = (int)$2; }
+    {
+      if ($2 > 0) {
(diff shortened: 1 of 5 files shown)


configparser.y
```diff
@@ -295,9 +295,25 @@ server_option:
   | VAR_IP_FREEBIND boolean
     { cfg_parser->opt->ip_freebind = $2; }
   | VAR_SEND_BUFFER_SIZE number
-    { cfg_parser->opt->send_buffer_size = (int)$2; }
+    {
+      if ($2 > 0) {
+        cfg_parser->opt->send_buffer_size = (int)$2;
+      } else if ($2 == 0) {
+        /* do nothing and use the default value */
```
wcawijngaards, line 302: If the user has multiple lines in config, the second zero would not use a default value. But that is unlikely.

wcawijngaards (approved):
The code looks nice.

For Unbound, the value of 0 means use the system default, that is without setting the setsockopt call. To keep the option backwards compatible 0 has to be the NSD default value, for NSD.

mozzieongit:
> For Unbound, the value of 0 means use the system default, that is without setting the setsockopt call. To keep the option backwards compatible 0 has to be the NSD default value, for NSD.

I don't quite understand what you mean to say.

wcawijngaards:
It would be nice if the config for Unbound and NSD are similar. The 0 value means something different today. 0 for Unbound means to use the system default buffer size. 0 for NSD means the NSD default buffer size. But I think it cannot be changed because of backwards compatibility.

PR #467, opened 2025-10-16
feat(xdp): add checksum verification on rx path

### LINKED ISSUE #465: AF_XDP sockets checksum verification
While looking at the AF_XDP experimental implementation, I noticed that checksums are not verified on the rx path. This means that, contrary to the "default" implementation, packets with an incorrect checksum will be processed as normal.

Since this implementation is still experimental, I am not entirely sure this would be considered an issue. I am mostly curious to understand if this behavior is known and/or intended.

RuiCunhaM:
This PR introduces checksum verification for IPv4 and UDP in the XDP rx path. It fixes #465 

For UDP I utilized the available functions in `xdp-util.h`. As for IP, I followed the example from [xdp-project](github.com/xdp-project/bpf-examples/blob/main/AF_XDP-example/xdpsock.c), and I "borrowed" the functions from the Kernel (I don't think this causes any issues in terms of Licensing).

Furthermore, since this implies some extra computing per each packet/request, I wanted to measure the performance implications. With a server running on a machine with an Intel i5-9500 and an Intel 10G X550T NIC, at a rate of 1M packets/requests per second, the average CPU usage across all CPU cores went from **11.18%** to **11.57%**.

For a single client sending synchronous requests, its rate drops from an average of **12360 req/s** to **12315 req/s**.

If this minor performance hit is an issue, we could perhaps add a compile flag so checksum verification would become optional.

### DIFF
--- a/xdp-server.c
+++ b/xdp-server.c
@@ -767,24 +767,43 @@ process_packet(struct xdp_server *xdp, uint8_t *pkt,
 		if (!(udp = parse_ipv6(ipv6)))
 			return -3;
 
+		uint16_t udp6_old_check = udp->check;
+		uint16_t udp6_check = calc_csum_udp6(udp, ipv6);
+
+		if (udp6_check != udp6_old_check)
+			return -4;
+
 		dnslen -= (uint32_t) (sizeof(*eth) + sizeof(*ipv6) + sizeof(*udp));
 		data_before_dnshdr_len += sizeof(*ipv6);
 
 		if (!dest_ip_allowed6(xdp, ipv6))
-			return -4;
+			return -5;
 
 		break;
 	} case ETH_P_IP: {
 		ipv4 = (struct iphdr *)(eth + 1);
 
+		__sum16 ipv4_old_check = ipv4->check;	
+		ipv4->check = 0;
+		ipv4->check = ip_fast_csum(ipv4, ipv4->ihl);
+
+		if (ipv4->check != ipv4_old_check)
+			return -6;
+
 		if (!(udp = parse_ipv4(ipv4)))
-			return -5;
+			return -7;
+
+		uint16_t udp4_old_check = udp->check;
+		uint16_t udp4_check = calc_csum_udp4(udp, ipv4);
+
+		if (udp4_check != udp4_old_check)
+			return -8;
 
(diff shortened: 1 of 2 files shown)


mozzieongit:
Thanks for the PR.
 
> (I don't think this causes any issues in terms of Licensing).

I'm not so sure about this statement. The file in the Linux kernel (`lib/checksum.c`) has the license `GPL-2.0-or-later`, we use the `BSD-3-Clause` for NSD. And roughly speaking, GPL is _infectious_, so we would have to use GPL as well when including GPL-licensed code in NSD. I'll ask, but to my knowledge, the `do_csum` function is sufficiently complex to be copyrightable and therefore we cannot simply copy that function without applying the GPL license. However, if the underlying algorithm is in the public domain or under a license compatible with NSD's license, then we could rewrite the function based on that algorithm.

> If this minor performance hit is an issue, we could perhaps add a compile flag so checksum verification would become optional.

Thanks for benchmarking this. The performance hit seems negligible. If anyone using this in the future would like to have a compile option for the checksums, we can provide it, but I'd wait for when they ask for it. No need to add it now.

RuiCunhaM:
> I'm not so sure about this statement. The file in the Linux kernel (lib/checksum.c) has the license GPL-2.0-or-later, we use the BSD-3-Clause for NSD. And roughly speaking, GPL is infectious, so we would have to use GPL as well when including GPL-licensed code in NSD. I'll ask, but to my knowledge, the do_csum function is sufficiently complex to be copyrightable and therefore we cannot simply copy that function without applying the GPL license. However, if the underlying algorithm is in the public domain or under a license compatible with NSD's license, then we could rewrite the function based on that algorithm.

That's a great point. I, in fact, misjudged this, sorry. Let me know if there is something I can help with.

mozzieongit:
A colleague mentioned that we might be able to use the FreeBSD kernel's checksum functions instead (https://cgit.freebsd.org/src/tree/sys/netinet/in_cksum.c). However, that file uses the BSD-**4**-Clause license, and would therefore require us to "display the following acknowledgement: This product includes software developed by the University of California, Berkeley and its contributors." for "All advertising materials mentioning features or use of this software", and I would have to ask if that is acceptable.
Looking for alternatives, maybe the tcpdump calculation (which has the BSD-3-Clause license) is fast enough (https://github.com/the-tcpdump-group/tcpdump/blob/master/in_cksum.c)?

PR #469, opened 2025-11-24
Add container build

jschlyter:
Add container (e.g., docker) build infrastructure. Should build, tag and publish container on release (`nsd:VERSION`) and on push to master (`nsd:latest`).


mozzieongit:
Hi, thanks for your contribution. The code/config look good.

However, providing container images for NSD is not something we want to maintain, currently. And we don't want to provide images now and then suddenly stop when something changes/breaks.

Your contribution would still be valuable to other users, so the whole `container` directory could be moved to the `contrib` directory and maintained by the community. The workflow would have to be removed.

jschlyter:
@mozzieongit I've removed the workflow and moved the `container` directory to `contrib`.

The solution to not breaking containers would be to always build them and fix them accordingly ;-)

mozzieongit:
> @mozzieongit I've removed the workflow and moved the `container` directory to `contrib`.

Thanks.

> The solution to not breaking containers would be to always build them and fix them accordingly ;-)

Certainly. It's just that _someone_ has to fix them.

PR #472, opened 2025-12-20
Reduce memory fragmentation with RRsets with many RRs in them

wtoorop:
Some zones, such as DNS-SD zones, have RRsets with many RRs in them. By minimizing the need to reallocate the list of RRs in an RRset, we will reduce memory fragmentation significantly for such zones. To this end, with this PR, RRs within an RRset, that are grouped together in a zone file, will be committed to the database in batches.


wtoorop:
@saj Do you think you can find the time to try out the [devel/reduce-fragmentation-with-many-RR-RRsets](/NLnetLabs/nsd/tree/devel/reduce-fragmentation-with-many-RR-RRsets) branch? Thanks!

saj:
@wtoorop oh sure it is the least I can do. :)  
I will try to post some results tomorrow -- or by Monday at the latest.

Thank you!

wtoorop:
> @wtoorop oh sure it is the least I can do. :) I will try to post some results tomorrow -- or by Monday at the latest.
> 
> Thank you!

Excellent and thank you!

saj:
https://lists.nlnetlabs.nl/pipermail/nsd-users/2025-December/003430.html

wcawijngaards (approved):
The code looks good. It reduces memory fragments for growing the RR array, instead of one at a time to a bunch. The nsd-mem changes look good too.

wtoorop:
I translated the approach used for loading zones in presentation format now also for processing transfers. Can you review again @wcawijngaards and @mozzieongit ? Thanks!

difffile.c
```diff
@@ -916,80 +1004,55 @@ add_RR(namedb_type* db, const dname_type* dname,
...
 	/* Now that the RR has been read with its RRtype specific read
 	 * routine, store the read data and rdata, for an ixfr store. */
-	if (ixfr_store)
+	if (ixfr_store) {
 		ixfr_store_addrr(ixfr_store, rr);
-
-	rrnum = find_rr_num(rrset, type, klass, rr, 1);
-	if(rrnum != -1) {
+	}
+	/* With the first RR for a RRset in this position in the zone file,
```
wcawijngaards: This is the process of a zone transfer, the comment could perhaps change.
```suggestion
	/* With the first RR for a RRset in this position in the zone transfer,
```

difffile.c
```diff
@@ -1336,12 +1408,13 @@ apply_ixfr(nsd_type* nsd, FILE *in, uint32_t serialno,
 		} else {
 			/* add this rr */
 			if(!(add_RR(nsd->db, owner, type, klass, ttl, packet,
-				rrlen, zone, softfail, ixfr_store))) {
+				rrlen, zone, softfail, ixfr_store, &collect_rrs))) {
```
wcawijngaards, line 1411: Perhaps add a comment to document, that if this does not exit the process, the domain table tree would be inconsistent.
```suggestion
				rrlen, zone, softfail, ixfr_store, &collect_rrs))) {
				/* The collect_rrs->rrs that have not been committed yet, need to
				   have rr_lower_usage(db, collect_rrs->rrs[i]); for them. This is
				   needed to remove references to the db domain tree. However
				   the failure is fatal to the process, so it is not necessary. */
```

zonec.c
```diff
@@ -413,6 +462,7 @@ zonec_read(
 	if (zone_parse(&parser, &options, &buffers, zonefile, &state) != 0) {
 		return state.errors;
```
wcawijngaards, line 473: Similar to the error case for zone transfers, if this fails, then the domain table tree would be in an inconsistent state. The error does not stop the process, it looks from `namedb_read_zonefile`, so it would need to clear the `state.c.rrs` with `rr_lower_usage` before returning out of function `zonec_read`.

wcawijngaards (approved):
The domain tree reference counts are wrong if the parser rejects an RR and stops the zone load, and the partial previous RRs in the collect_rrs are discarded, without calling rr_lower_usage on them.

The code looks fine, otherwise, and the fix for also the zone transfer case seems important.

wcawijngaards (approved):
The fix cleans up the database references and removes the memory usage, that looks good. The in depth fix for the fatal case looks good as well.

PR #477, opened 2026-03-06
Improve ignored old serial log message

tobbez:
Make it clear which serial is the old ignored one (received from the remote server) and which is the local, newer one.

You shouldn't have to read the source code to definitely tell which is which.

While you could infer which serial is which by checking which one is larger, it's not always easy to tell at a glance, so it's better to spell it out explicitly in the log message. I've already seen someone mix up the two serials in practice.

The best words I could come up with to describe the serials were `local`/`remote` which should make it clear which is which, but I'm open to better suggestions.


wcawijngaards (approved):
That looks nicer. Thank you for the improvement!

PR #481, opened 2026-04-22
Fix pedantic/CodeQL warning in sources

jaredmauch:
- udb.h: define enum udb_chunk_type before typedef; use [] flexible arrays.
- difffile.h: use flexible array for task_list_d.zname.
- rdata.c: drop stray }; after svcparam_must_not_have_value.
- verify.c: avoid preprocessor directives inside assert() arguments.
- xfrd-tcp.c: use j < 1 instead of j+1 <= 1 (same logic; CodeQL overflow rule).


mozzieongit:
Thanks for the PR

Issue #482, opened 2026-04-23
Consider using labels for zonestats in Prometheus metrics


ruuda:
In NSD 4.12.0, when you enable Prometheus metrics with `metrics-enable: yes`, and zonestats for say `example.com` and `example.org` with `zonestats: "example_com"` and `"example_org"` respectively, NSD will expose metrics in this format:

```
nsd_queries_by_type_total{type="A"} 198
nsd_queries_by_type_total{type="NS"} 198
...
nsd_zonestats_example_com_queries_by_type_total{type="A"} 99
nsd_zonestats_example_com_queries_by_type_total{type="NS"} 99
...
nsd_zonestats_example_org_queries_by_type_total{type="A"} 99
nsd_zonestats_example_org_queries_by_type_total{type="NS"} 99
...
```

The following format would be more useful:

```
nsd_zonestats_queries_by_type_total{type="A",zone="example.com"} 99
nsd_zonestats_queries_by_type_total{type="NS",zone="example.com"} 99
...
nsd_zonestats_queries_by_type_total{type="A",zone="example.org"} 99
nsd_zonestats_queries_by_type_total{type="NS",zone="example.org"} 99
...
```

In other words, put the zone in a label rather than in the metric name.

### Advantages

* You can use any zone name, it doesn’t need to be a valid Prometheus identifier. In particular, it may contain a dot.
* It unlocks Prometheus features that work over metrics with labels but not over separate metrics:
    - For example, we can select a subset of zones by regex, e.g. `nsd_zonestats_queries_by_type_total{zone =~ ".+\.nl"}`.
    - Functions like `sum`, `avg`, etc. become available.
* It makes it easier to graph metrics.
    * For example, you can query `rate(nsd_zonestats_queries_by_type_total{type="A"}[5m])` in the Prometheus webui, and it will graph the rates for all zones in a single graph.
    * This also applies to Grafana, you can use the zone label as series name, and then with a single query Grafana will all zones in a single graph. It is possible to achieve the same result with differently named metrics, but it requires manually duplicating the query `n_zones` times, and it requires manually updating the dashboard when zones are added or removed.
* Similar to graphing all zones at once, it is possible to create alerts for all zones at once, and use the `zone` label in the alert description.
* It doesn’t pollute Prometheus with `O(n_zones)` additional metrics. This matters for example in the Prometheus webui, or in graphing tools like Grafana, which autocomplete metric names, which is a valuable for exploring available metrics.

### Impact

[...]

PR #483, opened 2026-05-02
Prometheus metrics: Move zonestats from metric name to label

ruuda:
This implements the change proposed in #482, moving the zonestats name out of the metric name and into a label.

This pull request is structured in separate commits that can be reviewed one at a time.

* The first commit introduces a helper struct to format metrics with labels. This is only a refactor, no behavioral changes.
* It’s not really needed to use the helper everywhere because some metrics don’t need labels, but I thought it would be nicer for consistency, so in the second commit I switch all Prometheus metrics to use the helper.
   * Instead of `print_longnum`, I use a `PRIu64` format string, which is used in the codebase already, but only in one place. Does `print_longnum` exist only for historical reasons, or is it important to only use `%lu` in format strings?
   * Up to this point the output should not change, and I verified this by diffing the `/metrics` response on a binary built from `master` vs. this commit.
* Then move the zonestats name out of the prefix and into a label, which is now a very small change.
* I thought that instead of silently replacing disallowed characters in the zonestats name, it would be nicer to fail on it. I implemented that in the final commit, but it’s only validated at startup. It would be nice if `nsd-checkconf` could catch it already, but this is a bit tricky because the zonestats name can be a template like `%s` and we have to validate after applying the template, which looks like it would be a larger change to `nsd-checkconf`, so I’m not sure if that’s worth it.

I tested locally with Prometheus 3.11.2 that it can still parse the `/metrics` response, and it can.

### DIFF
--- a/metrics.h
+++ b/metrics.h
@@ -62,18 +62,18 @@ void daemon_metrics_attach(struct daemon_metrics* m, struct xfrd_state* xfrd);
 #ifdef BIND8_STATS
 
 /**
- * Check that a Prometheus label value contains no disallowed characters.
+ * Replace characters disallowed in Prometheus label values with '_'.
  *
  * According to [1], label values can be any UTF-8 characters, but backslash,
- * double-quote, and line feed must be escaped. We could do full validation but
- * in practice for zonestats names you don't need these characters anyway so we
- * just ban them.
+ * double-quote, and line feed must be escaped. We could escape them bad that
+ * changes the length of the string, and in practice for zonestats names you
+ * don't need these characters anyway so we just replace them.
  * [1]: https://prometheus.io/docs/instrumenting/exposition_formats/#comments-help-text-and-type-information>
  *
- * @param value: the label value to validate.
- * @return NULL on success or the failure reason on error.
+ * @param value: the label value to edit.
+ * @return 1 if any changes were needed, 0 if the value was already valid.
  */
-const char* metrics_validate_label_value(const char* value);
+int metrics_make_label_value_valid(char* value);
 
 /**
  * Print stats as prometheus metrics to HTTP buffer
--- a/metrics.h
+++ b/metrics.h
@@ -65,7 +65,7 @@ void daemon_metrics_attach(struct daemon_metrics* m, struct xfrd_state* xfrd);
  * Replace characters disallowed in Prometheus label values with '_'.
  *
  * According to [1], label values can be any UTF-8 characters, but backslash,
- * double-quote, and line feed must be escaped. We could escape them bad that
+ * double-quote, and line feed must be escaped. We could escape them but that
  * changes the length of the string, and in practice for zonestats names you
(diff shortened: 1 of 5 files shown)


options.c
```diff
@@ -2947,6 +2948,13 @@ unsigned getzonestatid(struct nsd_options* opt, struct zone_options* zopt)
 	if(!zopt->pattern->zonestats || zopt->pattern->zonestats[0]==0)
 		return 0; /* no zone stats */
 	statname = config_cook_string(zopt, zopt->pattern->zonestats);
+
+	const char* statname_error = metrics_validate_label_value(statname);
+	if (statname_error) {
+		log_msg(LOG_ERR, "invalid zonestats name \"%s\": %s",
+			statname, statname_error);
+		exit(1);
+	}
```
mozzieongit, line 2959: I see the appeal to verify on startup, but I don't think that it is a good idea. It will break existing configurations using zonestats when updating NSD even if they don't use Prometheus actively.

I would prefer to have the replacement of disallowed strings over the exit as before. However, you could still log a warning on startup when the zonestats name contains a disallowed string (and that is being replaced in the metrics output).

mozzieongit:
Thanks for the PR. I'll look at it the during the coming weeks, but cannot promise a timeline due to upcoming events

mozzieongit (changes requested):
Great work! Just the one note about the disallowed characters.

> I don't know why previously long numbers needed to be split, PRIu64 is
used in the codebase already (in one place, though PRIu32 and others
have more usage). Maybe because that code (at least the original in
remote.c) is very old?

Could be. I don't know either, but indeed replicated the behaviour from remote.c

options.c
ruuda: Makes sense. I implemented both in 910fa779.

ruuda: It does now log the warning twice, I suppose that happens because that code path is somehow called twice.

mozzieongit: Indeed. It is called for every instance of the `zonestats` option in the config file.

metrics.h
```diff
@@ -62,18 +62,18 @@ void daemon_metrics_attach(struct daemon_metrics* m, struct xfrd_state* xfrd);
...
 
 /**
- * Check that a Prometheus label value contains no disallowed characters.
+ * Replace characters disallowed in Prometheus label values with '_'.
  *
  * According to [1], label values can be any UTF-8 characters, but backslash,
- * double-quote, and line feed must be escaped. We could do full validation but
- * in practice for zonestats names you don't need these characters anyway so we
- * just ban them.
+ * double-quote, and line feed must be escaped. We could escape them bad that
```
mozzieongit:
```suggestion
 * double-quote, and line feed must be escaped. We could escape them but that
```

mozzieongit (approved):
Thanks a lot for the changes. I'll fix the typo and then ask the team if there is anything stopping a merge, otherwise I'll go ahead.

PR #484, opened 2026-05-09
OpenRC: fix network deps and support both supervisors

orlitzky:
1. Remove `need net` from the service dependencies. it's almost always wrong, and is in this case (the commit message explains why).
2. Support both start-stop-daemon _and_ supervise-daemon by moving the `supervisor=` line into the conf file. There are some things that s-s-d does that supervise-daemon still cannot do.

cc @lanodan


lanodan:
Nice, looks good to me

PR #485, opened 2026-05-14
Fix typo: Same as ip-address

simonho234:
This fixes a typo in nsd.conf.5.in where '\-' was incorrectly escaped. Change 'Same as ip\-address' to 'Same as ip-address'.


mozzieongit:
Hi,
this is not a typo, but deliberate. The escaped '-' gets rendered as a normal '-' in the man page.

PR #486, opened 2026-06-02
Correctly load tls-cert-bundle with chroot

bilias:
I've noticed an error about the `tls-cert-bundle`
`xfrd tls: Unable to set the certificate bundle file /var/nsd/etc/cert-bundle.pem`

I see that this comes from
```
server_start_xfrd()
  -> xfrd_init()
     -> xfrd_tcp_set_create()
         -> SSL_CTX_load_verify_locations()
```
which is happening after chroot

### DIFF
--- a/nsd.c
+++ b/nsd.c
@@ -1395,6 +1395,9 @@ main(int argc, char *argv[])
 		} else if (!file_inside_chroot(nsd.options->xfrdir, nsd.chrootdir)) {
 			error("xfrdir %s is not relative to %s: chroot not possible",
 				nsd.options->xfrdir, nsd.chrootdir);
+		} else if (!file_inside_chroot(nsd.options->tls_cert_bundle, nsd.chrootdir)) {
+			error("tls-cert-bundle %s is not relative to %s: chroot not possible",
+				nsd.options->tls_cert_bundle, nsd.chrootdir);
 		}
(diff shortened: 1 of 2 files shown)


bilias:
Added one more patch (https://github.com/NLnetLabs/nsd/pull/486/changes/99b6bf15287ff420741462c5ba5e113912a4f605)
If you have a zone with multiple acls on provide-xfr like this:

```
       provide-xfr: 1.2.3.4 tsig1 tls-auth-1
       provide-xfr: 2.3.4.5 tsig2 tls-auth-2
```
server is checking ACLs independently of tls-auth. 

So in this case, when secondary 2.3.4.5 connects with tls-auth for zone transfer,
acl checks first for IP and tsig key. After that we have the SSL check.
It fails because it searches first for `auth-domain-name` of `tls-auth-1`

Server should not check other tls-auth definitions.
If an IP is matched in a specific ACL, check that specific tls-auth that is defined in that specific ACL.

Effectively this is only a logic change (no code change) and we move the TLS AUTH check code inside 
`if(acl_addr_matches(acl, q) && acl_key_matches(acl, q)) {`
and not after it.

bilias:
New PR (https://github.com/NLnetLabs/nsd/pull/489)

Issue #363, opened 2024-07-29
Read client certificates before dropping privileges


k0ekk0ek:
Andreas Schulze gave some feedback on the mutual tls(-auth) (#185 and #337) feature on the [nsd-users](https://lists.nlnetlabs.nl/pipermail/nsd-users/2024-July/003326.html) mailing list.

It appears certificates specified with `client-cert` are read on opening a connection. Think about updating the logic to read the certificates before dropping privileges, a more extensive description is provided by Andreas below:

NSD serving as simple TLS server is configured with:
```
server:
	username: nsd
	ip-address: ::@853
	tls-service-key: /path/to/key.pem
	tls-service-pem: /path/to/cert+intermediate.pem
	tls-port: 853

	# since 4.10.1rc2
	ip-address: ::@1853
	tls-auth-port: 1853
	tls-auth-xfr-only: yes
	tls-cert-bundle: /path/to/ca-certificates.crt
```

In this mode, `/path/to/*.pem` may accessible for the root user only.

Now, when adding a tls-auth for the purpose of client authentication I add:
```
tls-auth:
	name: primary.nsd.example
	auth-domain-name: primary.nsd.example
	client-cert: /path/to/cert+intermediate.pem
	client-key: /path/to/key.pem
```
Here, the files `/path/to/*.pem` are used by a child process with limited privileges of the username 'nsd' It would be better, if nsd read all tls-auth client-[cert|key] data before dropping privileges. Then the files could be still limited to be readable by the root user.

PR #488, opened 2026-06-04
Prevent race on SSL CTX

bilias:
`xfrd_tcp_open()` currently loads the client certificate and private key into the global SSL_CTX (`set->ssl_ctx`) which is shared across all pipelines.

This causes a race condition when multiple zones have different tls-auth definitions with different client certificates. Concurrent calls to `xfrd_tcp_open()` can overwrite each other's certificate/key in the shared context, causing a pipeline to authenticate with the wrong identity or fail `SSL_check_private_key()` due to a cert/key mismatch.

Fix this by loading the private key/certificate directly into the per-pipeline SSL object (`tp->ssl`)

I've kept `SSL_CTX_set_default_passwd_cb` and `SSL_CTX_set_default_passwd_cb_userdata` on the global CTX for now since LibreSSL that I use does not have `SSL_set_default_passwd_cb` and `SSL_set_default_passwd_cb_userdata`. So we still have a race condition there that should be documented (use same password for all encrypted keys!)

Ideally this should be solved by loading tls-auth early on startup before dropping privileges (see #363) and before chroot while keeping a different SSL_CTX per tls-auth.

### DIFF
--- a/xfrd-tcp.c
+++ b/xfrd-tcp.c
@@ -995,24 +995,36 @@ xfrd_tcp_open(struct xfrd_tcp_set* set, struct xfrd_tcp_pipeline* tp,
 	if (zone->master->tls_auth_options &&
 		zone->master->tls_auth_options->auth_domain_name) {
 #ifdef HAVE_TLS_1_3
-		/* Load client certificate (if provided) */
+		// password callback is only available on global ctx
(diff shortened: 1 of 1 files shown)


bilias:
New PR (https://github.com/NLnetLabs/nsd/pull/490)

PR #492, opened 2026-06-29
Fix IXFR difference sequences detection in xfrd

wtoorop:
Detected by failed nsec3_resalt.tdir, nsec3_resaltbroken.tdir and testplan_ixfr.tdir


wcawijngaards (approved):
The code changes looks good. So that it can disambiguate between what happened just before when it performs the duplicate serial check.

wtoorop:
> The code changes looks good. So that it can disambiguate between what happened just before when it performs the duplicate serial check.

Thanks! That was really quick!

PR #494, opened 2026-07-02
EDNS0 padding option support with DoT

wtoorop:
Implementation + test package for RFC 7830,
with the recommendations from RFC 8467 (i.e. pad responses with block size 468)


query.c
```diff
@@ -1839,7 +1839,19 @@ query_add_optional(query_type *q, nsd_type *nsd, uint32_t *now_p)
...
+		if(q->edns.padding) {
+			size_t cur_sz = buffer_position(q->packet) + 2 + q->edns.opt_reserved_space;
+			size_t padded_sz = (((cur_sz - 1) / PADDING_BLOCK_SZ) + 1) * PADDING_BLOCK_SZ;
+			size_t to_padd = padded_sz - cur_sz;
+			/* Need 4 bytes for option code and length */
+			q->edns.padding = to_padd >= 4 ? to_padd
+			                : to_padd >  0 ? (PADDING_BLOCK_SZ + to_padd) 
+					: 0; /* Multiple of PADDING_BLOCK_SZ,
+			                      * so no outgoing padding option */
+			if(q->edns.padding) {
```
wcawijngaards, line 1853: The code suggested stops padding if it does not fit in the buffer, or 65535 the max TCP length for a message. It can catch that without the change, then it drops the EDNS option section. This may also be acceptable.

```suggestion
			if(!buffer_available(q->packet, 2+q->edns.opt_reserved_space+q->edns.padding) || cur_sz + q->edns.padding > 65535)
				q->edns.padding = 0;
			if(q->edns.padding) {
```

wcawijngaards (approved):
The code looks okay to me. I wanted to add a length and buffer capacity check when deciding on how much to pad. But the edns options section is already length checked, but then it can drop the pad but not other edns options.

query.c
wtoorop: I like that improvement, thanks!

PR #496, opened 2026-07-13
Fix allow-query syntax in catalog documentation

maertsen:


wcawijngaards (approved):
The fix looks good. Nice to improve the documentation for NSD.

PR #497, opened 2026-07-15
An -f option to nsd-checkzone to print all domain names fully qualified

wtoorop:


wcawijngaards (approved):
The code looks good.

wtoorop:
I'm closing, because it is not needed right now, and there might be a better way to do this in the future

PR #499, opened 2026-07-15
The docpath and pvd SvcParams

wtoorop:
This PR will have all the new SvcParams accumulated


wcawijngaards (approved):
The code looks okay. The simdzone changes are also okay. The pvd change has no commit or pullrequest in simdzone.

PR #511, opened 2026-09-09
Fix to retry a catalog consumer zone over AXFR if IXFR didn't apply

wtoorop:
The tpkg test uses an artificially induced failure to apply a xfr for a specific SOA serial (0xBADFEED). For this nsd needs to be compiled with TESTING_CODE defined. This can be enabled by `--enable-testing`. Maybe this comes in handy for future issues that are hard to reproduce (like for example the throttling the TCP receive window down to 1 with 51cc814)

### DIFF
--- a/tpkg/catzone-recovery.tdir/catzone-recovery.conf.in
+++ b/tpkg/catzone-recovery.tdir/catzone-recovery.conf.in
@@ -1,0 +1,24 @@
+server:
+	logfile: "nsd.log"
+	pidfile: "nsd.pid"
+	username: ""
+	zonesdir: ""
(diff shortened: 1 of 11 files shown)


wcawijngaards (approved):
The fix looks okay. The test framework is okay to have. Not sure if the changes in the main code are worth it, for the test case that is added, but it is good to check that the failover path reloads the consumer zone correctly.

wtoorop:
> Not sure if the changes in the main code are worth it, for the test case that is added

Ack. I'll put it aside for now then in a separate branch then to be used for test packages that need modified main code.

PR #514, opened 2026-09-15
A report-channel config option to return a Report-Channel EDNS option in responses

wtoorop:


wcawijngaards (approved):
The code looks okay. The test looks fine too.

